Issue #62, opened 2015-04-13
Could you please support "Debug Directories" sections?


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. dumpbin /headers MFCApplication1.exe
2. You can see "Debug Directories" section after "SECTION HEADER #3".
3. It will include the GUID, PDB path.
4. Currently pefile won't show this section.

What is the expected output? What do you see instead?
  Debug Directories

        Time Type       Size      RVA  Pointer
    -------- ------ -------- -------- --------
    54E40A41 cv           4A 000450A4    328A4    Format: RSDS, {23993D7C-78DD-4
C82-B73A-E49D3DB0CE40}, 1, d:\temp\MFCApplication1\Debug\MFCApplication1.pdb
    54E40A41 feat         14 000450F0    328F0    Counts: Pre-VC++ 11.00=0, C/C+
+=48, /GS=48, /sdl=11, reserved=0
pefile will not show this part.


What version of the product are you using? On what operating system?
vc 2013 dumpbin under win 8.1; pefile 1.2.10-139.

Please provide any additional information below.
Could you please support the "debug directories" sections?
I'd like to get the GUID and pdb path.
Thanks a lot.

```

Original issue reported on code.google.com by `jeffreyr...@gmail.com` on 4 Mar 2015 at 10:24

nyx0:
Update your pefile and try this:
```python
import struct
import pefile

pe = pefile.PE('foobar.exe')

for i in pe.DIRECTORY_ENTRY_DEBUG:               
    guid_data1 = struct.pack('<I', i.entries.Signature_Data1).encode('hex')
    guid_data2 = struct.pack('<H', i.entries.Signature_Data2).encode('hex')
    guid_data3 = struct.pack('<H', i.entries.Signature_Data3).encode('hex')
    guid_data4 = struct.pack('<Q', i.entries.Signature_Data4).encode('hex')
    pdb_path = i.entries.PdbFileName
print '{{{}-{}-{}-{}}}'.format(guid_data1, guid_data2, guid_data3, guid_data4)
print '{}'.format(pdb_path)
```

WHOLETTHEDOG-OUT:
It didn't work.
![image](https://user-images.githubusercontent.com/13184758/158352372-f6c9cb93-4f35-493b-9300-2a3edb2f0ed5.png)

PR #71, opened 2015-05-15
Change the repo url in setup.py

nyx0:
However there is an issue with:
```Python
__revision__ = "$LastChangedRevision$"
```
You can find an alternative here if you want https://packaging.python.org/en/latest/single_source_version.html#single-sourcing-the-version


nyx0:
b85677a should solve the issue #62

Issue #66, opened 2015-04-24
pefile not showing all imports 


frbapolkosnik:
A colleague of mine stepped across this dll that breaks pefile's import parsing.

Pefile shows only two imported dlls (msjet35.dll and kernel32.dll), but going through the structures USER32.dll and ADVAPI32.dll imports are missing,

MD5 is 954ceb4d7c7dc5e94ea237cf96d387a3

http://pedump.me/954ceb4d7c7dc5e94ea237cf96d387a3/

I think that VirusTotal is using pefile since it also shows the truncated imports:
https://www.virustotal.com/en/file/66c74e4c9dbd1d33b22f63cd0318b72dea88f9dbb4d36a3383d3da20b037d42e/analysis/

![msrd2x35_imports](https://cloud.githubusercontent.com/assets/4399806/7326087/c1316d20-ea90-11e4-9c94-4b4946dfab6f.gif)

frbapolkosnik:
pefile throws an error on the third import, and because of that it breaks out of the while loop, and stops processing the rest of the imports...

I've added some info to the error messages:
Error parsing the import directory. Invalid Import data at RVA: 0x503c (Invalid Import data at RVA: 0x5e03c (Invalid Import Table information. originalFirstThunk:0x5e178  FirstThunk:0x5e298 Both ILT and IAT appear to be broken.)

webstergd:
I am doing a port of PEInfo now. I will see if I can tackle this and push a patch to both systems

erocarrera:
pefile version 2016.3.4 fixes this issue.

PR #79, opened 2016-01-05
Fix for empty import table

Ma-Shell:
PEFILE now properly parses the import table. Previously the import table would return before all imports were processed. Thus, this Closes issue #66

Previously, the file referenced in this issue imports four libraries:
msjet35.dll, KERNEL32.dll, USER32.dll, ADVAPI32.dll. If the IAT and ILT of USER32.dll is empty (both directly point to locations containing 0), pefile will raise a PEFormatError with the text "Invalid Import Table information. Both ILT and IAT appear to be broken."
As a result, USER32.dll is not added to the imports and the imports from ADVAPI32.dll are not even looked into. 

In our studies this also effected additional files and would result with pefile reporting a subsect of the actually import table. 

To fix this behavior, this pull request returns an empty list, if both (the IAT and the ILT) have zero entries, instead of throwing an error or returning None.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -3678,7 +3678,7 @@ class PE:
                     'Invalid import data at RVA: 0x%x (%s)' % ( rva, e.value) )
                 break
 
-            if not import_data:
+            if import_data is None:
                 continue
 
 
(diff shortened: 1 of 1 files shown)


erocarrera:
This fix introduced problems handling files with a potential large number of invalid import table entries. I've fixed issue #66 in a more complete way.

Issue #59, opened 2015-04-13
peutils.py line 423


GoogleCodeExporter:
```
if value == '??' or value == '?0' :

should be

if '?' in value:

This allows (no parsing errors) on some of the signatures that look for various 
combinations of don't cares.
```

Original issue reported on code.google.com by `soos...@gmail.com` on 3 Dec 2014 at 10:47

erocarrera:
Fixed in develop branch.

PR #81, opened 2016-01-13
Fix issue #59

nyx0:


nyx0:
@erocarrera I think there is a issue with the check. Can you take a look?

erocarrera:
Thanks! I've addressed the without needing to merge.

PR #89, opened 2016-02-14
Use "python -m pefile" to get printed dump

techtonik:


erocarrera:
Neat! good idea, thanks!

techtonik:
np. Glad you like it.

PR #110, opened 2016-05-03
correct bug in get_imphash

sebdraven:
Hi,

I use pefile with python 3.5 and there a bug in the function get_imphash:

for entry in self.DIRECTORY_ENTRY_IMPORT:
            libname = entry.dll.lower()
            parts = libname.rsplit('.', 1)

libname is a byte array and '.' is str so rsplit makes an error: 

File "/usr/local/lib/python3.5/dist-packages/pefile-2016.3.28-py3.5.egg/pefile.py", line 3625, in get_imphash
TypeError: a bytes-like object is required, not 'str'

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -3622,7 +3622,7 @@ class PE(object):
             return ""
         for entry in self.DIRECTORY_ENTRY_IMPORT:
             libname = entry.dll.lower()
-            parts = libname.rsplit('.', 1)
+            parts = libname.rsplit(b'.', 1)
             if len(parts) > 1 and parts[1] in exts:
                 libname = parts[0]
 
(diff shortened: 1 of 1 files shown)


Rafiot:
Fails on python3.5:

```
Python 3.5.1+ (default, Mar 30 2016, 22:46:26) 
[GCC 5.3.1 20160330] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pefile
>>> p = pefile.PE("/home/raphael/gits/viper/misp_samples/2624/73.exe")
>>> p.get_imphash()
  File "/home/raphael/gits/pefile/pefile.py", line 3641, in get_imphash
    impstrs.append(b'%s.%s' % (libname.lower(),funcname.lower()))
TypeError: %b requires bytes, or an object that implements __bytes__, not 'str'

```

erocarrera:
I believe the merged issue 120 has taken care of this bug.

PR #116, opened 2016-05-09
Check, if dpg_type_partial is not None to avoid crash

leprovokateur:
It can happen the following while processing a dll:

AttributeError: 'NoneType' object has no attribute 'Unicode'

This is because __unpack_data__ can return None for whatever reason.


erocarrera:
Thanks for the fix. It's already included in pull request #109 which I've merged.

Issue #105, opened 2016-04-22
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 208687: ordinal not in range(128)


ibrahimkaraguzel:
Hello;
I am ok with, while building and installing pefile via Python2.7. I got No errors.

Problem starts while building Python3.4:
`Building source...
Sandbox enabled build...
running build
running build_py
creating build
creating build/lib
copying pefile.py -> build/lib
copying peutils.py -> build/lib
creating build/lib/ordlookup
copying ordlookup/__init__.py -> build/lib/ordlookup
copying ordlookup/oleaut32.py -> build/lib/ordlookup
copying ordlookup/ws2_32.py -> build/lib/ordlookup
warning: build_py: byte-compiling is disabled, skipping.

DEBUG: return value for "python setup.py build" is 0 ---->build with python2.7, no errors
Traceback (most recent call last):
  File "setup.py", line 57, in <module>
    pefile_version = _read_attr('__version__')
  File "setup.py", line 39, in _read_attr
    match = re.search(regex, f.read())
  File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 208687: ordinal not in range(128)
DEBUG: return value for "python3.4 setup.py build" is 1   --->error`
I can't fix this unicode error unfortunately. I am using linux.

techtonik:
Have you tried Python 3.5?

https://travis-ci.org/erocarrera/pefile/jobs/123192709

techtonik:
Can't confirm this on Linux. Attach exact version of pefile.py here, exact version of Linux and Python.

ibrahimkaraguzel:
Hello dude
I am developer at PisiLinux (www.pisilinux.org)
I downloaded this package of yours: https://github.com/erocarrera/pefile/archive/v2016.3.28.tar.gz
My Linux version is 4.1.19
My python3 version is 3.4.3

ibrahimkaraguzel:
Traceback (most recent call last):
  File "setup.py", line 57, in <module>
    pefile_version = _read_attr('__version__')
  File "setup.py", line 39, in _read_attr
    match = re.search(regex, f.read())
  File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
[...]

PR #126, opened 2016-06-02
Remove only unicode character from pefile.py

stephenrauch:
A lone unicode character (with a reasonable ascii replacement) is causing setup.py to choke on a machine using cp950. This is another fix for Issue #105. The change at PR #111 is likely a better long term fix, but this PR should also fix the problem as it makes pefile.py pure ascii.

Also seen at [pyinstaller #2019](https://github.com/pyinstaller/pyinstaller/issues/2019)


erocarrera:
Thanks, good catch!

escapewindow:
Could you make another release with this fix?

PR #133, opened 2016-06-27
Fix is_driver()

hillu:
Commit f5ff78606528d371714e50db567b92cb2eec31b2 introduced a regression:
Corrupted (truncated) files where no valid import section could
be found would cause an "has no attribute 'DIRECTORY_ENTRY_IMPORT'"
exception to be raised in is_driver().


erocarrera:
I've had to address the related issue of files where imports had not been parsed due to the PE having been parsed with "fast_load=True". In dealing with that I've also implemented your fix.

PR #140, opened 2016-09-09
Should fix #130

### LINKED ISSUE #130: dump_dict raises ValueError
``` python
  File "pefile.py", line 4574, in dump_dict
    dump_dict['DOS_HEADER'] = self.DOS_HEADER.dump_dict()
  File "pefile.py", line 943, in dump_dict
    val = bytes([b for b in val if b != 0])
  File "/home/user/.local/lib/python2.7/site-packages/future/types/newbytes.py", line 108, in __new__
    raise ValueError('bytes must be in range(0, 256)')
ValueError: bytes must be in range(0, 256)
```

nyx0:
fix issue #130

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -940,7 +940,7 @@ class Structure(object):
                         except exceptions.ValueError as e:
                             val = '0x%-8X [INVALID TIME]' % val
                 else:
-                    val = bytes([b for b in val if b != 0])
+                    val = ''.join(filter(lambda c:c != '\0', str(val)))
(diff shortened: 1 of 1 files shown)


knowmalware:
#161 is a different implementation of this fix.  Let me know what you think so we can close this or that to make it easier on the maintainer.

erocarrera:
As far as I can see issue #130 should have been addressed. I can't reproduce it anymore.

PR #143, opened 2016-09-28
Fixed corner case for when a dll only has export ordinals.

MalwareMechanic:
Some DLL's don't have export function names at all. In this case, the pefile code as it stands will bail out early when parsing the export directory due to not being able to get the section containing the `AddressOfNames`. This PR will allow it to continue parsing the export directory even if no export function names are present.


DoomMuffins:
Wouldn't it be better to check the `export_dir.NumberOfNames` field rather than checking whether `export_dir.AddressOfNames` is an RVA that points within a section?

PR #149, opened 2016-10-12
Fix for 2 python3 issues

quoscient:
Btw could you please tell me how to decrypt test_data.tar.bz2.enc so I can add testcases for the issues?


erocarrera:
Thanks for the PR!

PR #153, opened 2016-11-02
Fix some errors with cython

Drewsif:
Cython raised some errors when trying to compile the code so I fixed them. The wrong ValueError was being used and convert_to_printable() was missing so I created one.


knowmalware:
@Drewsif see #161 for a better convert_to_printable() fix

PR #161, opened 2017-01-07
Fixes #130, issues with dump_dict() on some files.

### LINKED ISSUE #130: dump_dict raises ValueError

knowmalware:
Fix a variable scope issue introduced when function named 'b' was created.
Fix resource strings by removing call to a function that no longer exists.
  This fix is more in line with the dump() implementation by using the
  same encode/decode sequence.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -966,7 +966,7 @@ class Structure(object):
                         except exceptions.ValueError as e:
                             val = '0x%-8X [INVALID TIME]' % val
                 else:
-                    val = b([b for b in val if b != 0])
+                    val = b([x for x in val if x != 0])
 
                 dump_dict[key] = {'FileOffset': self.__field_offsets__[key] + self.__file_offset__,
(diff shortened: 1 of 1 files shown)


erocarrera:
I believe the bug this PR tried to fix is no longer an issue. Yell at me otherwise :)

knowmalware:
/me yells

Still a problem with latest master.  This should still fix it.

PR #164, opened 2017-01-22
Add support for parsing and dumping COFF object files

zlatkok:
Add support for parsing and dumping COFF object files.
Dump symbol table and section relocations.
Fix some small issues.


bernhl:
@erocarrera are you generally fine with adding COFF parsing to pefile?
Since the PR has conflicts with master somebody would need to work on that anyway so I propose to move as much of the COFF code and constants as possible out from pefile.py into a separate coff.py.
In particular, I'd like to keep the original `__parse__` function and override it in the COFF class (which inherits from PE).

learn-more:
@erocarrera Any update on this?

PR #165, opened 2017-01-31
fix bugs encountered when porting packrat

songlh:


bernhl:
Can you please provide the sha256 of the input file?
@erocarrera do we want to get here https://github.com/erocarrera/pefile/blob/06991406248f232a25be152361ebdaa12853218c/pefile.py#L3227 
if `versioninfo_struct = None`? If not, the patch is good (maybe replace `if versioninfo_string == None` with `if versioninfo_string is None`)

PR #166, opened 2017-01-31
encoding/parsing error fixes

suicidejack:
We've been using pefile over a very large sample set of binaries (5+ million) and this PR addresses some of the issues that we have run across.

For versioninfo_string we discovered that sometimes that information could not be retrieved which would cause pefile to throw an exception in the subsequent logic.  That fix just skips that logic which doesn't apply when there is no version info.

For the symbol.name fix - sometimes that value would come through as a string and other times it comes through as bytes.  This just checks what the type of symbol.name is and makes sure that it is always a string.

### DIFF
--- a/ordlookup/__init__.py
+++ b/ordlookup/__init__.py
@@ -1,5 +1,5 @@
-import sys
 from __future__ import absolute_import
+import sys
 from . import ws2_32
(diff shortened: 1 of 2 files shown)


boddumanohar:
fixes #198

PR #175, opened 2017-04-03
Fix catch-all exception clause in parse_resources_directory

tdube:
This particular call to PE.get_data() was not in a try-catch block and caused an uncaught PEFormatError exception when parsing malware sample 4cf7e36c5ef5b37f3231740794fecdaec5635b176320288a50694ca5be7ac898 (32-bit PE). PE.get_data() attempts to read from RVA 0xffe93eb8 which is invalid.


bernhl:
I can verify both the issue and that it works fine with the patch.

tdube:
PR #166 already addressed most of this PR, but includes a catch-all exception clause. This PR now limits the exception to PEFormatError.

PR #176, opened 2017-04-09
Fix parsing of ordinal-only exports

yirkha:
Hi, I noticed a problem where pefile could not load exports for a DLL and tracked it down to it having ordinal-only exports, no named ones. A common example is mfc100u.dll or similar.

The attached patch fixes the issue.

Thanks for your consideration.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -3513,13 +3513,16 @@ class PE(object):
 
         max_failed_entries_before_giving_up = 10
 
-        section = self.get_section_by_rva(export_dir.AddressOfNames)
-        if not section:
-            self.__warnings.append(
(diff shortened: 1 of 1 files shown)


bernhl:
I tried to verify this issue for `f33603f89e29979a6045f749ce60ad036db1a3aaef74ee7b41a522e23ee36575 mfc140u.dll` and the export directory seems to load fine. Can you please verify the issue still exists with the latest version, post the sha256 of the file you have issues with and elaborate on the unexpected behavior?

yirkha:
Thanks for the follow-up.
You are right, it has been fixed already by #143 (9625bfaa).
Merged in 1 day after I had submitted this PR :smiley:
Closing.

Issue #171, opened 2017-02-23
Missing testfiles in tarball


shibumi:
Hello,
there are missing testfiles in the release tarballs. Could you please include them and maybe bump a new version if you have new features? Thanks!

stephenrauch:
Some of test files are malware and are encrypted,  See https://github.com/erocarrera/pefile/issues/82#issuecomment-192018385

pombredanne:
FYI to support the effort by @Rafiot to get  pefile to work better with Python 3 in @viper-framework , I committed this repo of no-encrypted test files for pefile https://github.com/viper-framework/pefile-tests  including a set of excellent warped PEs by @angea (some of them that make pefile choke alright were moved to the "failing" directory) and including some PE with undocumented origins (these will be documented eventually)

It would be best if @erocarrera you could get some of your test files made public .... (including for instance your highly coveted `empty_file` that surely does not deserve military-grade encryption ;) )

pombredanne:
@erocarrera is there anything special about the files that you have in your encrypted test set and that are referenced in these places:
 - MSVBVM60.DLL here and elsewhere https://github.com/erocarrera/pefile/blob/aa44f2d96d1a7e7e9c84ec994efdcc068d6f6e3f/tests/pefile_test.py#L140
- mfc40.dll and kernel32.dll at https://github.com/erocarrera/pefile/blob/aa44f2d96d1a7e7e9c84ec994efdcc068d6f6e3f/tests/pefile_test.py#L158
- cmd.exe here https://github.com/erocarrera/pefile/blob/aa44f2d96d1a7e7e9c84ec994efdcc068d6f6e3f/tests/pefile_test.py#L175 (that I verified externall to be a vanilla  cmd.exe from Win7 64.

Because if these are the problematic test files, there can easily be replaced with vanilla equivalent PEs. And @angea set of malformed PEs are otherwise great (and fairly small) 

Beyond these: 

- in https://github.com/erocarrera/pefile/blob/aa44f2d96d1a7e7e9c84ec994efdcc068d6f6e3f/tests/pefile_test.py#L171
66c74e4c9dbd1d33b22f63cd0318b72dea88f9dbb4d36a3383d3da20b037d42e is likely a sha256 for a version of msrd2x35.dll at https://virustotal.com/en/file/66c74e4c9dbd1d33b22f63cd0318b72dea88f9dbb4d36a3383d3da20b037d42e/analysis/

[...]

PR #180, opened 2017-05-03
Fixed minor typo

pombredanne:
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -1046,7 +1046,7 @@ class SectionStructure(Structure):
         if name == 'Characteristics':
             section_flags = retrieve_flags(SECTION_CHARACTERISTICS, 'IMAGE_SCN_')
 
-            # Set the section's flags according the the Characteristics member
+            # Set the section's flags according to the Characteristics member
             set_flags(self, val, section_flags)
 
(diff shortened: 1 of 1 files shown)


pombredanne:
@erocarrera any reason why the tests crash on just a comment typo fix? It looks like some of the test files are encrypted? what would this be? that sounds weird for an open source project...

erocarrera:
Hi Philippe, thanks for the fix. The test should not fail because of this change. I've run them locally and I can verify there's no issue with them. I'm not entirely sure why they failed to run in Travis.
And you're right, the tests are encrypted, some of the test files can't be shared, and others are malware, and I can't have them laying publicly in the clear.
I've merged the PR, thanks again!

pombredanne:
@erocarrera because I could not run tests locally short of having you encryption key, I created a whole test suite from scratch at https://github.com/viper-framework/pefile-tests together with @Rafiot ...
This could be something you could consider and I can send a PR your way if need be.
Some of it is "reversed-engineered" from you test code some is new addition for weird PEs ... so things should be pretty close or may be even improved when compared to your own test suite files. See https://github.com/erocarrera/pefile/issues/171#issuecomment-299266627

PR #181, opened 2017-05-05
fix bug of str and bytes comparison in python3

ShingYang:
fix bug of str and bytes comparison in python3, this change applies to python 2 as well


Rafiot:
If you use python3, I strongly recommend you to have a look at this commit: https://github.com/erocarrera/pefile/pull/189
It covers a lot of cases.

ShingYang:
since the issue fixed in #189 , will close the pull request.

PR #195, opened 2017-06-30
Change typecheck is is_valid_function_name to use type() instead of isinstance()

rhelmot:
This is a performance consideration: with this change, the time for loading kernel32.dll goes from 4 seconds to 1 second.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -1450,7 +1450,7 @@ else:
         string.digits + b'_?@$()<>')
 
 def is_valid_function_name(s):
-    if s is None or not isinstance(s, (str, bytes)):
+    if s is None or type(s) not in (str, bytes):
         return False
     for c in set(s):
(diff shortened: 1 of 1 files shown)


rhelmot:
Never mind, it looks like this breaks a lot of things. I don't really know why and that upsets me a lot, but I'll figure this out.

I did notice through profiling, though, that an absurd amount of time was spent inside isinstance calls, so _something_ like this would probably be helpful, applied somewhere else with maybe some more care. 😒

rhelmot:
NEVER MIND NEVER MIND

I was testing on the version on pypi. The version on master has apparently had everything fixed, so the four hours of optimization I just sunk into this project aren't worth shit 😒 😒 😒 😒 

Can a maintainer please upload the current release to pypi?

PR #205, opened 2017-08-14
Remove past.builtins.old_div

tlecomte:
Dear @erocarrera this pull request is about the import of the "past" library. Importing past has bad side-effects for projects that depend on pefile, such as pyinstaller. Using past introduces a dependency on the test module, which in turn depends on other modules... In the end, when freezing a Python program with PyInstaller, this forces many modules to be included for little benefit.

In pefile, there is now just one use of past left, and it looks like using past is not needed since the code needs float division anyway. This pull request removes that use.

Thanks for considering this request!

### DIFF
--- a/peutils.py
+++ b/peutils.py
@@ -13,7 +13,6 @@ from __future__ import division
 from future import standard_library
 standard_library.install_aliases()
 from builtins import range
-from past.utils import old_div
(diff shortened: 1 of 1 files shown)


erocarrera:
Thank you!

PR #206, opened 2017-08-18
Use VirtualSize field to zero-pad loaded segments

rhelmot:
This fixes an emulation issue I was having in an application using pefile.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -1001,32 +1001,35 @@ class SectionStructure(Structure):
         Returns bytes() under Python 3.x and set() under 2.7
         """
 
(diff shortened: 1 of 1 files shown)


rhelmot:
I just pushed another, larger commit that expands this behavior and fixes more emulation issues I was having - it appears the windows loader will happily accept PE files with sections whose SizeOfRawData extends past the end of the file itself, by padding the remainder with zeros. In order to handle this while also protecting against maliciously crafted executables, I added a length cap to the load by using the max_virtual_address variable already present in get_memory_mapped_image.

PR #216, opened 2017-11-30
Avoid to raise an exception when the type of the _s_ variable is bytes or bytearray

enkomio:
In function _is_valid_function_name_ and _is_valid_dos_filename_ there is a wrong type reference during the loop.

The problematic code is:

    for c in set(s):
        if c not in allowed_function_name:
            return False

if the _s_ variable is of type _bytes_ or _bytearray_ an exception is raised during the iteration. This can be easily fixed by iterating on the string value of the _s_ variable.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -1432,7 +1432,7 @@ def is_valid_dos_filename(s):
         return False
     # Allow path separators as import names can contain directories.
     allowed = allowed_filename + b'\\/'
-    for c in set(s):
+    for c in set(str(s)):
         if c not in allowed:
             return False
     return True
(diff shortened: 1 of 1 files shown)


erocarrera:
I get an Exception with your proposed modifications running with Python3:

  File "pefile/tests/pefile_test.py", line 144, in test_selective_loading_integrity
    pe.parse_data_directories(directories= list(range(0x10)))
  File "./pefile.py", line 2464, in parse_data_directories
    value = entry[1](dir_entry.VirtualAddress, dir_entry.Size)
  File "./pefile.py", line 3882, in parse_import_directory
    max_length = max_len)
  File "./pefile.py", line 4016, in parse_imports
    if not is_valid_function_name(imp_name):
  File "./pefile.py", line 1458, in is_valid_function_name
    if c not in allowed_function_name:
TypeError: a bytes-like object is required, not 'str'

enkomio:
I have replaced the fix with a version which depends on the python version. The call to the _canonicalize_ function will modify the value only in python 2.

PR #222, opened 2018-01-11
issue#108 solution - relocate_image() modification

smart-rabbit:
Modifications description:
1. to check the presence of reloc directory we use 'self.OPTIONAL_HEADER.DATA_DIRECTORY[5].Size', not 'hasattr(self, 'DIRECTORY_ENTRY_BASERELOC')', because reloc directory can be unloaded(unparsed). [loc 5303]
2. if there is no reloc directory exception raise (for first time commented) [locs 5420-5421]
3. if reloc directory is present and  unloaded(unparsed) - its loading(parsing) added [locs 5304-5305]
4. 'self.OPTIONAL_HEADER.ImageBase' correction added to correspond new Image Base [loc 5382]
5. correction of VAs(virtual addresses) occurrences in directories information added [locs 5384-5418]

P.S. locs 5389-5396 and 5415-5418 can be deleted - i leave them to show that this directories was analyzed,if they contain VAs(virtual addresses) occurrences.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -5300,7 +5300,9 @@ class PE(object):
 
         relocation_difference = new_ImageBase - self.OPTIONAL_HEADER.ImageBase
 
-        if hasattr(self, 'DIRECTORY_ENTRY_BASERELOC'):
+        if self.OPTIONAL_HEADER.DATA_DIRECTORY[5].Size:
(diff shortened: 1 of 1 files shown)


erocarrera:
Thanks!

PR #254, opened 2019-02-18
Improve PE size / overlay start calculations

### LINKED ISSUE #253: Improve PE size / overlay start calculations
There are a few cases where calling trim() results in an executable that's missing expected data or where calling get_overlay_data_start_offset() incorrectly calculates certain data as being overlay data.

The ones I've observed include:
 - The Authenticode signature being treated as overlay data (See: https://github.com/erocarrera/pefile/issues/104#issuecomment-429037686)
 - An executable having a COFF symbol table / string table
 - An executable with section headers but no data after them

This affects scripts that use pefile to perform embedded EXE extraction (it seems to be used frequently for this purpose).

recvfrom:
Implements fixes for #253 

Make it so that the following are handled better:
 - EXEs with Authenticode signatures
 - EXEs with a COFF symbol table / string table
 - EXEs with section headers but no data after

I can contribute tests and/or test binaries too - what's the preferred way of doing this, given that the test data appears encrypted?

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -5608,19 +5608,46 @@ class PE(object):
                 return offset_and_size
             return largest_offset_and_size
 
+        # Factor in the offset / size of the COFF Symbol Table if one exists
+        if (self.FILE_HEADER.PointerToSymbolTable != 0):
+
+            # The COFF Symbol Table size must be computed by multiplying
+            # the number of symbols by the size of the symbol table entry
+            # structure
+            coff_sym_table_size = self.FILE_HEADER.NumberOfSymbols * 18
+
+            # The COFF String Table follows the Symbol Table, and at a minimum
+            # contains a 4-byte size value (set to 4 if there are no strings in
+            # the table)
+            coff_str_table_size = self.get_dword_from_offset(
+                self.FILE_HEADER.PointerToSymbolTable + coff_sym_table_size)
+
+            if coff_str_table_size is not None:
+                largest_offset_and_size = update_if_sum_is_larger_and_within_file(
+                    (self.FILE_HEADER.PointerToSymbolTable, coff_sym_table_size + coff_str_table_size))
+
         if hasattr(self, 'OPTIONAL_HEADER'):
             largest_offset_and_size = update_if_sum_is_larger_and_within_file(
                 (self.OPTIONAL_HEADER.get_file_offset(), self.FILE_HEADER.SizeOfOptionalHeader))
+
+            # SizeOfHeaders stores the size of the MZ, PE, and section hdrs,
+            # rounded up to a multiple of the FileAlignment
+            largest_offset_and_size = update_if_sum_is_larger_and_within_file(
+                (0, self.OPTIONAL_HEADER.SizeOfHeaders))
 
         for section in self.sections:
             largest_offset_and_size = update_if_sum_is_larger_and_within_file(
                 (section.PointerToRawData, section.SizeOfRawData))
 
-        skip_directories = [DIRECTORY_ENTRY['IMAGE_DIRECTORY_ENTRY_SECURITY']]
-
         for idx, directory in enumerate(self.OPTIONAL_HEADER.DATA_DIRECTORY):
-            if idx in skip_directories:
+
+            if idx == DIRECTORY_ENTRY['IMAGE_DIRECTORY_ENTRY_SECURITY']:
+                # For the Security / Certificate Table, the VirtualAddress is
+                # a file offset instead of an RVA
+                largest_offset_and_size = update_if_sum_is_larger_and_within_file(
+                    (directory.VirtualAddress, directory.Size))
                 continue
+
             try:
                 largest_offset_and_size = update_if_sum_is_larger_and_within_file(
                     (self.get_offset_from_rva(directory.VirtualAddress), directory.Size))


erocarrera:
Hi,
If the test binaries are in VirusTotal, just give me the hashes and I'll get them from there. Otherwise you can send the binaries directly to me and I'll bundle them into the tests.

pefile.py
```diff
@@ -5608,19 +5608,46 @@ def update_if_sum_is_larger_and_within_file(offset_and_size, file_size=len(self.
...
-        skip_directories = [DIRECTORY_ENTRY['IMAGE_DIRECTORY_ENTRY_SECURITY']]
-
         for idx, directory in enumerate(self.OPTIONAL_HEADER.DATA_DIRECTORY):
-            if idx in skip_directories:
+
+            if idx == DIRECTORY_ENTRY['IMAGE_DIRECTORY_ENTRY_SECURITY']:
+                # For the Security / Certificate Table, the VirtualAddress is
+                # a file offset instead of an RVA
+                largest_offset_and_size = update_if_sum_is_larger_and_within_file(
+                    (directory.VirtualAddress, directory.Size))
```
erocarrera: The first item in the tuple given to `update_if_sum_is_larger_and_within_file()` should be a file offset, not a virtual address.

recvfrom: The comment above the call is meant to clarify this - for IMAGE_DIRECTORY_ENTRY_SECURITY, a file offset is stored in the VirtualAddress field instead of an actual Relative Virtual Address like in the other DIRECTORY_ENTRYs.  For reference, see: https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format#the-attribute-certificate-table-image-only .

recvfrom:
Here are two examples for each case:

EXEs with Authenticode signatures
 - 00048c246c8db3c309b759631057f1a5704296803a2ba23e0d9530d14986a130
 - 001a26ff51bf6babf6325983f512cf8d84cadee1ca36f166a41702d94c1b0841

EXEs with a COFF symbol table / string table
 - 01794f55fab26842c12e2a67fc218ad9c1a9201ccf0bf2fbd9f5815d6f20182f
 - 03d896e59d78d4338cae141ea52447190fe9ebd6362acd16d4cd2954039ed5d7

EXEs with section headers but no data after
 - 76c13fe37652df8ce2fde315b6eae4d2e0ec7f9424b4a0d6fc661354a7679da0
 - 39d41e1814a82488f14acfb06e96920ca5a633dc90acb31b21ab98b3cebdef5b

pefile.py
j-t-1: `IMAGE_DIRECTORY_ENTRY_SECURITY` is the security certificate, which is the same thing as Authenticode?

The loader verifies `IMAGE_DIRECTORY_ENTRY_SECURITY` and if valid the PE file is loaded into memory, but excluding `IMAGE_DIRECTORY_ENTRY_SECURITY` which has essentially done its job. So the `IMAGE_DIRECTORY_ENTRY_SECURITY` is a somewhat fuzzy entity, being part of the PE file (processed by the loader, which other overlays are not) but also still being at the end of the file (usually) and not loaded into memory.

This part of the PR could be correct.  Need to read documentation to confirm that the security certificate is not an overlay. Complicating this, I think some tools treat it as an overlay.

recvfrom:
@nightlark If you have access to VirusTotal all the files in my earlier comment should be there. I don't have access anymore, and if you don't either I can ask around to see if there's someone that might be able to pull them from there to add to pefile-tests

nightlark:
Unfortunately I don't have access to VirusTotal. I've been looking into if there's a way for "researchers" to get access for downloading samples without paying an absurd fee, and potentially access through work; neither has made much progress.

erocarrera:
I still have access to VT. I downloaded the six files above:
[issue 254.zip](https://github.com/user-attachments/files/26927027/issue.254.zip)

j-t-1:
The loader verifies `IMAGE_DIRECTORY_ENTRY_SECURITY` and if valid the PE file is loaded into memory but excluding `IMAGE_DIRECTORY_ENTRY_SECURITY` which has essentially done its job. The `IMAGE_DIRECTORY_ENTRY_SECURITY` is a somewhat fuzzy entity, being part of the PE file (processed by the loader, which other overlays are not) but also still being at the end of the file (usually) and not loaded into memory.

This PR excludes the certificate from the overlay. A new parameter could be added `exclude_certificate` that allows the user to include/exclude the certificate in the overlay. `PE.get_overlay` and `PE.trim` would also take the new parameter. The default value could be `True`. This could be done after this PR.

nightlark:
Is the loader excluding that image directory entry something that is seen in the Windows research kernel by chance?

If the certificate is excluded from the overlay, do we have some other way for users to access the data?

pefile.py
```diff
@@ -7851,16 +7869,25 @@ def update_if_sum_is_larger_and_within_file(
...
 
-        skip_directories = [DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_SECURITY"]]
-
         for idx, directory in enumerate(self.OPTIONAL_HEADER.DATA_DIRECTORY):
-            if idx in skip_directories:
+
+            if idx == DIRECTORY_ENTRY['IMAGE_DIRECTORY_ENTRY_SECURITY']:
+                # For the Security / Certificate Table, the VirtualAddress is
+                # a file offset instead of an RVA
+                largest_offset_and_size = update_if_sum_is_larger_and_within_file(
```
nightlark, line 7887: For an additional layer of clarity here, perhaps we should create a temporary variable named file_offset that just gets set to the directory.VirtualAddress to use in this function call.

j-t-1:
> Is the loader excluding that image directory entry something that is seen in the Windows research kernel by chance?

Do you know the function or module that may contain this? There is a `LdrVerifyMappedImageMatchesChecksum` that could be useful for checking `‎‎PE.generate_checksum‎` (not directly relevant to this PR) but I only found the declaration and not the definition.

> If the certificate is excluded from the overlay, do we have some other way for users to access the data?

No. `PE.get_overlay` and `PE.trim` use `PE.get_overlay_data_start_offset`, and so we need to add `include_certificate` to all of them. There still would not be an API to get only the certificate if the overlay had a certificate and data afterwards. If this was wanted, a new function `PE.get_certificate` would keep `PE.get_overlay` and `PE.trim` simple.

PR #258, opened 2019-03-28
Decode bytes object before iterating

siliconsheep:
Iterating over a bytes or bytearray object returns integers representing Unicode characters, so those should be decoded to strings first to get the expected behavior.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -1439,7 +1439,14 @@ def is_valid_dos_filename(s):
         return False
     # Allow path separators as import names can contain directories.
     allowed = allowed_filename + b'\\/'
-    for c in set(s):
+    if isinstance(s, (bytes, bytearray)):
+        try:
+            _s = s.decode(sys.getdefaultencoding())
+        except UnicodeDecodeError:
+            return False
+    else:
+        _s = s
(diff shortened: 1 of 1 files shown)


pefile.py
```diff
@@ -1439,7 +1439,14 @@ def is_valid_dos_filename(s):
         return False
     # Allow path separators as import names can contain directories.
     allowed = allowed_filename + b'\\/'
-    for c in set(s):
+    if isinstance(s, (bytes, bytearray)):
+        try:
+            _s = s.decode(sys.getdefaultencoding())
```
JayFoxRox: I don't understand why?

This *is* supposed to check for a custom byte-wise encoding. Hence `allowed_filename` is also a byte-string, and not a human-readable-string.

---

Doing as you suggest here breaks the algorithm and adds a lot of variance as different systems might use different encodings.

Instead, the check in line 1438 might have to be changed to *not* accept strings, or to decode them (which the `set()` might do, I'm not sure).

erocarrera: I would like to understand what issues are the objective of this change. Can you provide some examples of cases where the PR improves the output of pefile?

PR #263, opened 2019-06-15
Remove print. Fix format. Improve error handling.

nicholastoddsmith:
Malformed COFF symbol tables can cause pefile to crash (see 690EB2BFE038A2632F50E8A7E2FB9F64, 3E14E435D4B01127B8526F4B985347E3, A6763BAF63AF7467B3AD3486A0785C9F, etc). Apparently the DWORD symbol table size is bogus and this causes pefile to index out of bound. This commit adds additional error handling to allow parsing to continue even if this occurs.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -1093,7 +1093,7 @@ class Structure(object):
                     if key.startswith('Signature_'):
                         val_str = '{:8X}'.format(val)
                     else:
-                        val_str = '{#:8X}'.format(val)
+                        val_str = '0x{:8X}'.format(val)
 
                     if key == 'TimeDateStamp' or key == 'dwTimeStamp':
                         try:
@@ -2322,7 +2322,7 @@ class PE(object):
                     'AddressOfEntryPoint lies outside the sections\' boundaries. '
                     'AddressOfEntryPoint: 0x%x' %
                     self.OPTIONAL_HEADER.AddressOfEntryPoint )
-                
+
         if not fast_load:
             self.full_load()
 
@@ -2517,6 +2517,7 @@ class PE(object):
 
             reloc_size = Structure(self.__COFF_RELOCATION_format__).sizeof()
 
+            has_sym = hasattr(self, 'symbols')
             for i in range(section.NumberOfRelocations):
                 start = section.PointerToRelocations + reloc_size * i
                 end = start + reloc_size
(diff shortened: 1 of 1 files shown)


pefile.py
```diff
@@ -2660,21 +2661,25 @@ def parse_symbol_table(self):
...
 
-            if self.string_table_offset + 4 >= len(self.__data__):
+            self.symbols = []
+            self.strings = {}
+
+            if (self.string_table_offset + 4) >= len(self.__data__):
                 self.__warnings.append('Symbol table is corrupt')
                 return
 
             string_table_size = struct.unpack('<I', self.__data__[self.string_table_offset : self.string_table_offset + 4])[0]
```
nicholastoddsmith: This is the DWORD value that can apparently be invalid in some PE files. Loaded one of the samples in IDA pro and it complained of the same value for the string value size.

pefile.py
```diff
@@ -2660,21 +2661,25 @@ def parse_symbol_table(self):
...
 
-            if string_table_size < 4:
+            # This can be an invalid offset triggering an index error later
+            offset = self.string_table_offset + string_table_size - 1
+            if (string_table_size < 4) or (offset >= len(self.__data__)):
                 self.__warnings.append('Invalid string table size')
                 return
 
-            if self.__data__[self.string_table_offset + string_table_size - 1] != 0:
```
nicholastoddsmith: This line was crashing.

pefile.py
```diff
@@ -2697,7 +2702,6 @@ def get_symbol_name(symbol, string_table, string_table_size):
                             name = s
                         else:
                             end = string_table[offset:].find(0)
-                            print(offset, end)
```
nicholastoddsmith: My script was producing a lot of output in the most recent version of _pefile_ and I identified this as the line producing it.

pefile.py
```diff
@@ -2713,7 +2717,6 @@ def get_symbol_name(symbol, string_table, string_table_size):
 
                 return name
 
-            self.symbols = []
```
nicholastoddsmith: This value is reference elsewhere and can cause _AttributeError_ if it's not defined.

pefile.py
```diff
@@ -5503,13 +5506,16 @@ def dump_section_relocations(self, section):
         lines.append('-' * 79)
         reloc_table = MACHINE_TYPE_TO_RELOCATION.get(self.FILE_HEADER.Machine)
 
+        if not hasattr(self, 'symbols'):
+            lines.append('<PARSE ERROR>')
+            return lines
```
nicholastoddsmith: This is to avoid dump_info from crashing if the COFF parsing was aborted.

Issue #264, opened 2019-07-03
DATA_DIRECTORY missing entries for malformed values of NumberOfRvaAndSizes


huettenhain:
I noticed during malware analysis that `pefile` could better handle certain cases where the value of `NumberOfRvaAndSizes` has been reduced to hide the .NET directory information from certain parsers. Such a change can be detected by comparing the `SizeOfOptionalHeader` field with the size of the optional header if the given `NumberOfRvaAndSizes`  is assumed.

To reproduce the problem, take any .NET binary and reduce the value of `NumberOfRvaAndSizes` with a hex editor to `0x0000000E`, say. I took a copy of `RegAsm.exe` and patched it:
```
>>> data = open('RegAsm.Patched.exe', 'rb').read()
>>> pe = PE(data=data)
>>> pprint(pe.OPTIONAL_HEADER.DATA_DIRECTORY)
[<Structure: [IMAGE_DIRECTORY_ENTRY_EXPORT] 0xF8 0x0 VirtualAddress: 0x0 0xFC 0x4 Size: 0x0>,
 <Structure: [IMAGE_DIRECTORY_ENTRY_IMPORT] 0x100 0x0 VirtualAddress: 0xB78C 0x104 0x4 Size: 0x4F>,
 <Structure: [IMAGE_DIRECTORY_ENTRY_RESOURCE] 0x108 0x0 VirtualAddress: 0xC000 0x10C 0x4 Size: 0x608>,
 <Structure: [IMAGE_DIRECTORY_ENTRY_EXCEPTION] 0x110 0x0 VirtualAddress: 0x0 0x114 0x4 Size: 0x0>,
 <Structure: [IMAGE_DIRECTORY_ENTRY_SECURITY] 0x118 0x0 VirtualAddress: 0x0 0x11C 0x4 Size: 0x0>,
 <Structure: [IMAGE_DIRECTORY_ENTRY_BASERELOC] 0x120 0x0 VirtualAddress: 0xE000 0x124 0x4 Size: 0xC>,
 <Structure: [IMAGE_DIRECTORY_ENTRY_DEBUG] 0x128 0x0 VirtualAddress: 0xB720 0x12C 0x4 Size: 0x1C>,
 <Structure: [IMAGE_DIRECTORY_ENTRY_COPYRIGHT] 0x130 0x0 VirtualAddress: 0x0 0x134 0x4 Size: 0x0>,
[...]

PR #265, opened 2019-07-03
handling of too small NumberOfRvaAndSizes values

huettenhain:
This is a proposed fix for #264.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -2216,6 +2216,15 @@ class PE(object):
             self.NT_HEADERS.FILE_HEADER = self.FILE_HEADER
             self.NT_HEADERS.OPTIONAL_HEADER = self.OPTIONAL_HEADER
 
+            # Detect artificially reduced values of NumberOfRvaAndSizes
+            #
+            directory_delta = (self.FILE_HEADER.SizeOfOptionalHeader 
+                - (self.OPTIONAL_HEADER.sizeof() + self.OPTIONAL_HEADER.NumberOfRvaAndSizes * 8)) // 8
+            if directory_delta > 0:
+                self.__warnings.append(
+                    'SizeofOptionalHeader indicates that NumberOfRvaAndSizes is off by at least %d.' % directory_delta)
+                self.OPTIONAL_HEADER.NumberOfRvaAndSizes += directory_delta
+
             # Windows 8 specific check
             #
             if self.OPTIONAL_HEADER.AddressOfEntryPoint <  self.OPTIONAL_HEADER.SizeOfHeaders:


pefile.py
```diff
@@ -2216,6 +2216,15 @@ def __parse__(self, fname, data, fast_load, is_obj):
             self.NT_HEADERS.FILE_HEADER = self.FILE_HEADER
             self.NT_HEADERS.OPTIONAL_HEADER = self.OPTIONAL_HEADER
 
+            # Detect artificially reduced values of NumberOfRvaAndSizes
+            #
+            directory_delta = (self.FILE_HEADER.SizeOfOptionalHeader
```
nightlark, line 2221: I actually just came across a sample where SizeOfOptionalHeader is also inaccurate -- the optional header (including the data directories array) is clearly 240 (0xf0), but then SizeOfOptionalHeader is 160 (0xa0).

10 of the data directory entries are "missing" (e.g. the data directory array entry is present, but the address and size are both == 0), which is communicated by NumberOfRvaAndSizes being set to 6 (e.g. just a convenient precalculated count of "missing" entries). 10*8 bytes = 80 bytes (or 0x50), which just happens to be the difference between the actual space taken by the optional header and the SizeOfOptionalHeader value (0xf0 - 0xa0 = 0x50).

nightlark: Actually, I think the check you have is right as far as detecting misleading NumberOfRvaAndSizes. Upon closer inspection of the file, I think the tools we were using were interpreting data after the optional header as data directory entries.

huettenhain: Yep, this is pretty much the conclusion I also reached =).

huettenhain:
I have rebased the PR onto the most recent version since there has been some recent support. I'd still like to see this in pefile.

pefile.py
```diff
@@ -3245,6 +3245,15 @@ def __parse__(self, fname, data, fast_load):
         self.NT_HEADERS.FILE_HEADER = self.FILE_HEADER
         self.NT_HEADERS.OPTIONAL_HEADER = self.OPTIONAL_HEADER
 
+        # Detect artificially reduced values of NumberOfRvaAndSizes
+        #
+        directory_delta = (self.FILE_HEADER.SizeOfOptionalHeader 
+            - (self.OPTIONAL_HEADER.sizeof() + self.OPTIONAL_HEADER.NumberOfRvaAndSizes * 8)) // 8
+        if directory_delta > 0:
+            self.__warnings.append(
+                'SizeofOptionalHeader indicates that NumberOfRvaAndSizes is off by at least %d.' % directory_delta)
```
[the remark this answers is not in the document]
huettenhain: Done via bf8cf08; note that the rest of pefile does not use f-strings, so now it does stand out a bit ;-).

nightlark:
It's fine if source and destination are both master, since they are in different repositories -- this one and huettenhain's fork.

PR #269, opened 2019-08-21
Fix format typo in Structure.dump()

laomaiweng:
`'{#:8X}'.format(val)` yields a `KeyError` exception complaining about the #.

Alternate-form # format specifier should come after the colon, so this PR fixes this which seems to be a mere typo.


malwarefrank:
I suggest changing "8" to "10" in that one line, since the prefix (0X) will use two characters.  That would also makes it align better in the output.

PR #274, opened 2019-11-08
Add Structure method to determine field size

shizmob:
This PR adds the method `Structure.sizeof_field()` to determine the size of a single field within the structure. I've found this useful to avoid hardcoding certain assumptions, especially with PE32 vs PE32+ in mind.


nightlark:
I believe I got this updated to work with the latest master branch. It looks like it does change some function signatures (return values from `set_format` and `set_bitfields_format`), which based on their names are part of the public API, making this a breaking change. I'm not sure how frequently these functions are actually used by pefile users though.

I'm not finding other code on GitHub that calls set_format and references pefile, other than repositories that just have a copy of pefile.py in them. I'd like to get thoughts from a few other people on if they think this potential breaking change seems okay or not.

j-t-1 (commented):
Seems okay.

This comment puts it well: https://github.com/erocarrera/pefile/pull/274#issuecomment-4063428585. The checks took under ten minutes, so nothing stands out. A call graph may help, clarifying where the functions containing these classes are instantiated.

j-t-1:
> I believe I got this updated to work with the latest master branch. It looks like it does change some function signatures (return values from `set_format` and `set_bitfields_format`), which based on their names are part of the public API, making this a breaking change. I'm not sure how frequently these functions are actually used by pefile users though.
> 
> I'm not finding other code on GitHub that calls set_format and references pefile, other than repositories that just have a copy of pefile.py in them. I'd like to get thoughts from a few other people on if they think this potential breaking change seems okay or not.

I think a breaking change here is okay and needed:
#595 removes `__unpacked_data_elms__` from `set_format` and is also a breaking change.

There may be other improvements around this (outside of this PR).

j-t-1:
@Auscitte would there be benefit in also adding `__field_sizes__` to `set_bitfields_format`?

Issue #228, opened 2018-03-28
pe.TimeDateStamp and /Brepro flag


0x00ach:
Hi,

Just for information: if a pe.DIRECTORY_ENTRY_DEBUG struct has a Type of 0x10, then windbg preview considers pe.TimeDateStamp as a build ID, not a valid timestamp. VS2013 with /Brepro just sets the TimeDateStamp to 0xFFFFFFFF and does not add the entry, but MS pe files have it (if you change an existing entry type to 0x10 of a regular pe file, then windbg preview adds the warning).

I dont really know how (and if) you can render this information in pefile, but it may be useful for malware analysts/tools : build IDs may in some cases seem "valid" and I did not found any other way to differentiate them.

Cheers

ladislav-zezula:
This is from `explorer.exe` v 10.0.17134.165: There are three IMAGE_DEBUG_DIRECTORY entries:
```
  00 00 00 00  ....        (00000000) - Characteristics
  F8 A9 31 40  ..1@        (4031A9F8) - TimeDateStamp (2004-02-17 06-43-20)
  00 00        ..          (    0000) - MajorVersion
  00 00        ..          (    0000) - MinorVersion
  02 00 00 00  ....        (00000002) - Type
  25 00 00 00  %...        (00000025) - SizeOfData
  10 76 23 00  .v#.        (00237610) - AddressOfRawData
  10 56 23 00  .V#.        (00235610) - PointerToRawData
--------------------------------------------------------------------------------
  6F 72 65 72  orer        (7265726F) - Characteristics
  2E 70 64 62  .pdb        (6264702E) - TimeDateStamp (2022-04-23 23-31-26)
  00 00        ..          (    0000) - MajorVersion
  00 00        ..          (    0000) - MinorVersion
  0D 00 00 00  ....        (0000000D) - Type
  78 07 00 00  x...        (00000778) - SizeOfData
  38 76 23 00  8v#.        (00237638) - AddressOfRawData
  38 56 23 00  8V#.        (00235638) - PointerToRawData
--------------------------------------------------------------------------------
  00 00 00 00  ....        (00000000) - Characteristics
  F8 A9 31 40  ..1@        (4031A9F8) - TimeDateStamp (2004-02-17 06-43-20)
  00 00        ..          (    0000) - MajorVersion
  00 00        ..          (    0000) - MinorVersion
  10 00 00 00  ....        (00000010) - Type
  00 00 00 00  ....        (00000000) - SizeOfData
[...]

PR #277, opened 2020-01-12
Recognize two new PE debug types

nirbheek:
commit 3f8ffa7181c7c4c5958a55b786bc5eac88289131:

```
Recognize the IMAGE_DEBUG_TYPE_REPRO debug type
The presence of this flag indicates that the PE binary was built with
`/Brepro` by the MSVC linker. See:
```
https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#debug-type

Closes https://github.com/erocarrera/pefile/issues/228

commit d3d26d204d613f483103d6b8cf15e03dcd72ee41:

```
Recognize EX_DLLCHARACTERISTICS debug type
Merely updates the two-way dict, but more work is likely needed to
properly support this.
```

https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#extended-dll-characteristics

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -210,7 +210,8 @@ debug_types = [
     ('IMAGE_DEBUG_TYPE_VC_FEATURE',     12),
     ('IMAGE_DEBUG_TYPE_POGO',           13),
     ('IMAGE_DEBUG_TYPE_ILTCG',          14),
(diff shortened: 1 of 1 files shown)


nirbheek:
Not really sure why the CI is failing, I didn't change anything that seems related :)

erocarrera:
Thanks!

PR #280, opened 2020-04-02
Added new type machine arm64.

xiva-wgt:
Added new type machine arm64. https://docs.microsoft.com/ru-ru/windows/win32/debug/pe-format?redirectedfrom=MSDN#machine-types


erocarrera:
Thanks!

PR #283, opened 2020-04-28
Fix crash formatting a warning message

teskje:
The warning message for export directories that contain more than 10 repeated entries uses the `{:s}` formatter for a byte value. Under Python 3, this crashes with the error `TypeError: unsupported format string passed to bytes.__format__`.

This commit fixes that crash by removing the `s` formatter. Without specifying that it works under both Python 2 and 3.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -3672,7 +3672,7 @@ class PE(object):
             if symbol_counts[(symbol_name, symbol_address)] > 10:
                 self.__warnings.append(
                     'Export directory contains more than 10 repeated entries '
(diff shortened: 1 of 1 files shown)


ra-kete:
Looks like all the PRs are currently failing the checks due to:

```
Encrypted environment variables have been removed for security reasons.

See https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions
```

This prevents the test data from being extracted, so tests that require it fail.

I think the easiest fix would be to just skip those tests if the test data files are not available.

PR #287, opened 2020-06-09
patch to FileInfo attribute access without proper validation

mlodic:
Hi @erocarrera! Thank you for your tool

While I was using `pefile` I found this little bug that causes some analysis to crash.

I noticed that you already fixed it in another parts in the code. The last one was missing.

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -1,0 +1,26 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


pombredanne:
@mlodic I guess you ended up pushing this  https://pypi.org/project/pefile-fork/ short of having this merged and a new release here?
I am tempted to do the same to  get the latest published too.

@erocarrera ping, do you plan to make a new release and merge some of the bug fixes in PRs? It's OK if you cannot  but ... just telling that you will or won't would be nice and enough so I can adjust accordingly.

mlodic:
I confirm. Then yes, I should have not pushed those extra commits here

erocarrera:
This had been fixed when I'd addressed #313.

PR #299, opened 2020-10-30
Ignore device driver extension for imphash

rakuy0:
Like when a PE imports GetPrinter out of winspool.drv.


erocarrera:
While the rationale for the change is good, its impact would be that imphash values would change for any files with imports from `drv` files. Unless I'm overlooking some other positives, I don't think it's worth adding the change despite its aesthetic value.

PR #306, opened 2021-01-12
Fix duplicate export detection for ordinals

learn-more:
`MFC42u.dll` has a lot of ordinals all pointing to the same address (around 90).
Since the ordinal number was not taken into account in the duplicate detection, this would reject the rest of the exports.


learn-more:
This is not correct.

Issue #318, opened 2021-05-14
prevent installation of latest version on Python 2.7


marunguy:
Windows 10 20H2 64bit, python 2.7.18.

In 2021.5.13 version,  python 2.7 support is stopped.
When I tried to install it on Python 2.7 I got the error below.
I think that the previous version (2019.4.18) must be installed instead of the 2021.5.13 version.
```
py -2.7-32 -m pip install -U pefile
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support pip 21.0 will remove support for this functionality.
Collecting pefile
  Using cached pefile-2021.5.13.tar.gz (66 kB)
    ERROR: Command errored out with exit status 1:
     command: 'D:\devtool\py27x86\python.exe' -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'d:\\users\\marunguy\\appdata\\local\\temp\\pip-install-ntlal5\\pefile\\setup.py'"'"'; __file__='"'"'d:\\users\\marunguy\\appdata\\local\\temp\\pip-install-ntlal5\\pefile\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base 'd:\users\marunguy\appdata\local\temp\pip-pip-egg-info-e2ku0w'
         cwd: d:\users\marunguy\appdata\local\temp\pip-install-ntlal5\pefile\
    Complete output (12 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "d:\users\marunguy\appdata\local\temp\pip-install-ntlal5\pefile\setup.py", line 86, in <module>
        long_description = "\n".join(_read_doc().split('\n')),
      File "d:\users\marunguy\appdata\local\temp\pip-install-ntlal5\pefile\setup.py", line 30, in _read_doc
        tree = ast.parse(f.read())
      File "D:\devtool\py27x86\lib\ast.py", line 37, in parse
        return compile(source, filename, mode, PyCF_ONLY_AST)
      File "<unknown>", line 3789
[...]

PR #319, opened 2021-05-17
Fix f-strings formatting to fix 318

digitalsleuth:
This fixes the f-strings interpolation at lines 3789 and 3790 which are only compatible in Python 3.6+. Now allows installation and usage for python2.7. 

Tested on Ubuntu 18.04 and 20.04.


digitalsleuth:
Proposed fix for #318.

PR #328, opened 2021-07-29
Include license metadata in setup.py

dieresys:
Having it here simplifies showing this information in pypi, with pip show, or parsing it with tools like pip-licenses.


erocarrera:
Thanks!

PR #329, opened 2021-08-17
Ignore padding in section.get_data()

daladim:
Hi,

sections have two sizes: VirtualSize and SizeOfRawData.
These may have a padding in the PE file (due to alignement constraints), but this padding will be ignored by the loader. In this case (and unless the caller explicitely asks for this padding through the `length` argument), we may assume that the caller is not interested in this padding, and we could truncate it.


erocarrera:
There are files containing sections with `VirtualSize = 0` and `SizeOfRawData > 0`, with this change `pefile` would not return any data for those sections.

daladim:
Hmm. What's the point of such values? I would guess that the data in these sections are never used (unless they are some kind of data overlays)?
In case the loader never loads such sections in memory, their data is never used and I would say that returning `[]` makes sense (and there is still the option of requesting the whole section anyway using a suitable `length` argument)

Maybe I'm wrong, and returning the data on file still looks the best default behaviour in your opinion?

daladim:
Or maybe, if you're not comfortable with changing a default behaviour, we could provide two functions instead of the current `get_data`: `get_data_including_padding` and `get_data_with_padding_trimmed` (or any rewording of such names)

daladim:
All right. Do you want to close this pull request then, or do you want me to add a line in the comment of this function that would "warn" about this?

erocarrera:
Hi daladim, 

If you mean a comment to `get_data` detailing a bit what we have discussed in this conversation, that sounds like a good idea. I am a bit queasy about modifying the default behavior of `get_data` as it would alter the behavior of lots of code using `pefile` (and I prefer the "greedy" behavior), however if you also want to add an additional argument to `get_data` or add a method that replicates more closely what the Windows loader does, I don't object to that. Maybe `get_data` could take a `behavior="default"/"loader"` keyword argument, where `"loader"` could implement your suggestion? Or maybe `padding=True/False`?

daladim:
Applying your suggestions, I have added an optional `ignore_padding` keyword, and documented it in the docstring.

erocarrera:
Thanks!

PR #331, opened 2021-08-29
Adding function to get rich header hash

ForensicITGuy:
Added the function `get_rich_header_hash()` which returns the md5 hash of the PE Rich Header, which can then be used with VT or other searches

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -16,28 +16,18 @@ gracefully.
 Copyright (c) 2005-2021 Ero Carrera <ero.carrera@gmail.com>
 """
 
-from __future__ import division
-from __future__ import print_function
-from builtins import bytes
-from builtins import chr
-from builtins import object
-from builtins import range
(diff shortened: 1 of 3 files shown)


erocarrera:
Hi, I see how this may come handy, thanks!
Given that md5 is on the way out and that pefile already imports sha1/sha256/sha512 in addition to md5 I think it'd be great if all were supported. Maybe you could add a family of methods like `get_rich_header_md5()` and likewise for the other supported hashes? alternatively `get_rich_header_hash()` could default to md5 but take an optional argument specifying the hash algorithm.

ForensicITGuy:
Hey @erocarrera that's an awesome idea, I hadn't even thought of it!

I'll get working on that in the next couple of days and should have something on it soon

ForensicITGuy:
Gotta close this PR b/c I somehow messed up my branch, will open another

PR #333, opened 2021-09-02
Adding Rich Header Hashing via multiple algorithms

ForensicITGuy:
Added the function get_rich_header_hash() which returns the hash of the PE Rich Header, which can then be used with VT or other searches.

The function supports multiple hashing algorithms per @erocarrera 's input here: https://github.com/erocarrera/pefile/pull/331#issuecomment-909347591

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -5272,6 +5272,25 @@ class PE:
 
         return import_descs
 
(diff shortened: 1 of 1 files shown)


erocarrera:
Neat, thanks! thanks for adding this functionality.

PR #339, opened 2021-12-27
Fix handling of pefile.PE(fast_load=False)

rokm:
The `fast_load` argument value in `pefile.PE()` constructor needs to be explicitly checked for being `None`. The existing truthy/falsy check means that `fast_load=False` always falls back to the global value (which, when set to `True`, causes a fast load even though user explicitly requested a full load).


its0x08 (approved):

erocarrera:
Thanks!

PR #340, opened 2022-02-20
Update PE to allow with statements

dinateper:
added __enter__ and __exit__ to PE to use with context managers


erocarrera:
Thanks!

PR #344, opened 2022-03-14
Speed up relocation process

elicn:
Relocation can take up to several minutes for large files (e.g. 64-bit version of `kernelbase.dll`).

The reason is largely because `__data__` stores contents as `bytes`, which does not have a `__setitem__` method. When data needs to be modified it uses slicing - which casues the entire content to be copied over and over and slow down the entire relocation process. This patch simply turns `__data__` into a `bytearray` and have it leverage its `__setitem__` method to modify its contents much faster.

Note: after analyzing the slowdown rootcause and writing the patch, I noticed this issue was already reported and had a PR submitted to fix it almost 3 years ago (#266). I don't know why that PR was not accepted, but hopefully a fresh one would do.

Note: this patch does not maintain compatibility to Python2

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -6962,13 +6962,17 @@ class PE:
             raise TypeError("data should be of type: bytes")
 
         if 0 <= offset < len(self.__data__):
-            self.__data__ = (
-                self.__data__[:offset] + data + self.__data__[offset + len(data) :]
-            )
+            self.set_data_bytes(offset, data)
         else:
             return False
 
         return True
+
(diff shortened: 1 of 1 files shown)


erocarrera:
Hi @elicn. Sorry for the delay, I don't get to work on pefile very regularly. The PR is great and the speed up very significant! I had overlooked @266, thanks for also bringing it to my attention.

elicn:
Great, thanks!

psrok1:
Thank you as well!

Issue #348, opened 2022-06-23
Remove future from dependencies


FantasqueX:
I noticed pefile requires python version >= 3.6. As a result, future is not needed. Is it possible to remove future from dependencies?

erocarrera:
That's quite possibly true! I will do so. Thanks for noticing.

JasperJuergensen:
@erocarrera I noticed, that the last release was more than half a year ago. The change fixing this Issue here, is still not included in a release. In the meantime a CVE has been published for the future package (CVE-2022-40899). It probably does not effect pefile, but for anyone with a dependency vulnerability management it would be nice, if there would be a new release without the future package as dependency.

themanifold:
Please publish this to pypi.

PR #349, opened 2022-06-26
Remove future from dependencies

FantasqueX:
Fix: #348

### DIFF
--- a/ordlookup/__init__.py
+++ b/ordlookup/__init__.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
-import sys
 from . import ws2_32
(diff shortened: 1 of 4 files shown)


mungojam:
@erocarrera could you please produce a release with this change in? There is a High Severity CVE reported against the `future` package but it doesn't appear to be maintained anymore so it just needs removing as a dependency like you have done here.

erocarrera:
This is now out as [release 2023.2.7 ](https://github.com/erocarrera/pefile/releases/tag/v2023.2.7)

doomedraven:
Thank you

PR #350, opened 2022-07-05
get_overlay_data_start_offset update to not skip offsets larger than filesize

0xatinder:
… returning result, otherwise if largest offset is more than filesize that will be skipped and some other smaller size_offset will be considered largest_size_offset resulting in wrong overlay offset within data refered to by headers

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -7308,7 +7308,8 @@ class PE:
         def update_if_sum_is_larger_and_within_file(
             offset_and_size, file_size=len(self.__data__)
         ):
-            if sum(offset_and_size) > sum(largest_offset_and_size):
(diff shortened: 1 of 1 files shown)


0xatinder:
merging

erocarrera:
The code in this PR modified the functionality of `update_if_sum_is_larger_and_within_file()` in an incompatible way. The file you point to appears to be corrupt/truncated. Sections are either bigger than the total file size or lay beyond the end of the actual file. Hence the overlay calculation falls back to a small value. I am not aware of a standardized way of computing the overlay in the case of PEs like the one you point to, if there is, please reopen the issue and let me know.

PR #352, opened 2022-08-31
Exclude INIT section from packed executable warning

omry99:
Many (non-packed) drivers have a INIT section with the IMAGE_SCN_MEM_WRITE + IMAGE_SCN_MEM_EXECUTE flags.
Therefore, it is not an indication of a packed executable and should be excluded from the warnings.

This can be demonstrated with this code:
```python
import pefile
from pathlib import Path

x=0
for path in Path(r"C:\Windows\System32").rglob('*.sys'):
    pe = pefile.PE(path, fast_load=True)
    for section in pe.sections:
        if section.IMAGE_SCN_MEM_WRITE and section.IMAGE_SCN_MEM_EXECUTE:
            if section.Name.rstrip(b"\x00") == b"INIT" and pe.is_driver():
                x+=1
                print(path.name)

print(f"Number of drivers (falsly) suspected as packed: {x}")
```

Side note: I have not found any evidence supporting the current exclusion (for the PAGE section).


nightlark:
I'm not sure what's going on with the one regression test that is failing due to a missing `next_section_virtual_address` attribute.

pefile.py
```diff
@@ -3682,8 +3682,8 @@ def parse_sections(self, offset, max_offset=0x10000000):
             if section.__dict__.get(
                 "IMAGE_SCN_MEM_WRITE", False
             ) and section.__dict__.get("IMAGE_SCN_MEM_EXECUTE", False):
-                if section.Name.rstrip(b"\x00") == b"PAGE" and self.is_driver():
-                    # Drivers can have a PAGE section with those flags set without
+                if section.Name.rstrip(b"\x00") in [b"PAGE", b"INIT"] and self.is_driver():
```
j-t-1:
```suggestion
                if section.Name.rstrip(b"\x00") in (b"PAGE", b"INIT") and self.is_driver():
```

nightlark:
One of the regression tests will need updating. However, I would like to run the provided sample code on Windows drivers just to verify how many false positives there with actual drivers.

pefile.py
```diff
@@ -3659,8 +3659,8 @@ def parse_sections(self, offset, max_offset=0x10000000):
                 getattr(section, "IMAGE_SCN_MEM_WRITE", False)
                 and getattr(section, "IMAGE_SCN_MEM_EXECUTE", False)
             ):
-                if section.Name.rstrip(b"\x00") == b"PAGE" and self.is_driver():
-                    # Drivers can have a PAGE section with those flags set without
+                if section.Name.rstrip(b"\x00") in (b"PAGE", b"INIT") and self.is_driver():
+                    # Drivers can have a PAGE/INIT section with those flags set without
```
j-t-1, line 3663:
```suggestion
                    # Drivers can have a PAGE/INIT section with those flags set without being malicious
```

pefile.py
```diff
@@ -3659,8 +3659,8 @@ def parse_sections(self, offset, max_offset=0x10000000):
...
                     # implying that it is malicious
```
j-t-1, line 3664:
```suggestion
                    # https://devblogs.microsoft.com/oldnewthing/20160901-00/?p=94225
```

j-t-1:
Once agreed, I will remove the parsing warning from tests/lief-samples-PE/PE64_x86-64_atapi.sys.dmp:

Suspicious flags set for section 4. Both IMAGE_SCN_MEM_WRITE and IMAGE_SCN_MEM_EXECUTE are set. This might indicate a packed executable.

> Side note: I have not found any evidence supporting the current exclusion (for the PAGE section).

@omry99 you may have already tested this as well, changing to just INIT does not give any regressions (apart from the expected change to PE64_x86-64_atapi.sys.dmp).
`if section.Name.rstrip(b"\x00") == b"INIT" and self.is_driver():`

This was the commit that added the existing check:
https://github.com/erocarrera/pefile/commit/53eb94929d15984cc5ce6d832f91e284bb82eb7a

A DriverEntry routine can be pageable and should be in an INIT segment so it will be discarded.
https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/writing-a-driverentry-routine

It seems slightly unlikely that PAGE alone would allow violation of W^X; this could be something that used to be allowed but has been tightened since.

@nightlark shall I merge this PR and open an issue about the exclusion of the PAGE section?

PR #360, opened 2022-11-15
Loosen export symbol validation:

learn-more:
- Allow more symbols in export names
- Allow exports with an RVA of '0'

This was noticed with the '%' symbol in exports from a delphi .bpl file.
The rest of the symbols were obtained by simply trying all (readable) ascii chars in a .def file, and linking with VS 2022's link.exe
Everything that did not produce a valid symbol was eliminated.


learn-more:
@erocarrera is there anything I can do to help get this merged?

learn-more:
Thanks!

erocarrera:
Thanks for the PR, I made a change so that the updated loosened validation only applies when parsing exports and not when parsing imports.

Issue #303, opened 2020-11-24
Addition of IMAGE_DLLCHARACTERISTICS_EX_CET_COMPAT


jxb5151:
Hello,

Microsoft has introduced the attribute [CETCOMPAT](https://docs.microsoft.com/en-us/cpp/build/reference/cetcompat?view=msvc-160&viewFallbackFrom=vs-2019) to mark compatibility with [Shadow Stack](https://software.intel.com/sites/default/files/managed/4d/2a/control-flow-enforcement-technology-preview.pdf) (page 10 if interested). 

Traditionally one would expect this to be captured in `DllCharacteristics`, but I believe those flags have been consumed. Now it looks like things are being placed in `IMAGE_DEBUG_TYPE_EX_DLLCHARACTERISTICS`. Could you please add `IMAGE_DLLCHARACTERISTICS_EX_CET_COMPAT` to the enum listing? 

Thank you for your time and work on this project.

PR #365, opened 2023-02-06
Add parsing for IMAGE_DEBUG_TYPE_EX_DLLCHARACTERISTICS

aursulis:
@erocarrera I've had a go at doing something for #303. This parses the `IMAGE_DEBUG_TYPE_EX_DLLCHARACTERISTICS` debug directory entry type according to what I've seen from self-built and Microsoft's own PE files. Their documentation at https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#debug-type is a bit lacking so I've had to invent names for fields etc.

What this doesn't do is pretty print the flags from `dump_info`. I have another commit for that here: https://github.com/aursulis/pefile/commit/0ca80ed35d6034aa71a727287e24f3bc69203e72 however looking at the full output it is really out of place.

Let me know what you think.


erocarrera:
This is great, thanks!

PR #367, opened 2023-02-19
Close the __data__ attribute before reassigning it

### LINKED ISSUE #356: generate_checksum() does not close __data__ before reassigning it
I am using `pefile` 2022.5.30 on Windows. Consider the following code, which corrects the header checksum of a PE file.
```Python
import pefile

with pefile.PE('libffi-8.dll') as file:
    file.OPTIONAL_HEADER.CheckSum = file.generate_checksum()
    contents = file.write()
with open('libffi-8.dll', 'wb') as file:
    file.write(contents)
```
With CPython 3.10.7, this code runs without error. With PyPy3.9, I get the following error.
```
  File "sandbox0.py", line 6, in <module>
    with open('libffi-8.dll', 'wb') as file:
OSError: [Errno 22] Invalid argument: 'libffi-8.dll'
```
The error happens because the first line of `generate_checksum()` fails to close `self.__data__` before reassigning it to a different object. If `self.__data__` is a memory-mapped file, then this seems to keep a lock on the file so that it cannot be opened in later code.
```Python
def generate_checksum(self):
    self.__data__ = self.write()
    ....
```
With CPython, no error happens because the reference-counting garbage collector takes care of the closing the memory-mapped file, while PyPy experiences an error because its garbage collector does not close the memory-mapped file immediately. We should not be relying on any particular garbage collection implementation. If I change the beginning of `generate_checksum()` to the following, then I no longer see the issue.
```Python
def generate_checksum(self):
    data = self.write()
    self.close()
    self.__data__ = data
```

adang1345:
This fixes #356 where calling `generate_checksum()` or `set_data_bytes()` will leave behind a lock on the PE file if we're running PyPy.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,6 +19,7 @@ jobs:
           - "3.8"
           - "3.9"
           - "3.10"
+          - "pypy3.9"
         include:
           - os: ubuntu-latest
             python-version: "3.10"
--- a/pefile.py
+++ b/pefile.py
@@ -2903,7 +2903,7 @@ class PE:
     def __exit__(self, type, value, traceback):
         self.close()
 
-    def close(self):
+    def _close_data(self):
         if (
             self.__from_file is True
             and hasattr(self, "__data__")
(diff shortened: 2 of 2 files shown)


erocarrera:
Thank you!

PR #377, opened 2023-07-07
Exphash from sha256 to md5 to match imphash

N0fix:
PR #354 added exphash using the `sha256` function instead of `md5`, which creates two issues:
- There is a lack of consistency between imphash (which uses md5) and exphash (currently using sha256).
- [The associated YARA implementation (ongoing PR) uses md5](https://github.com/VirusTotal/yara/pull/1795).

This PR proposes using `md5` instead of `sha256` for exphash to maintain consistency.

Note: I believe this confusion stems from a discrepancy between the original blog post about exphash, [which mentions the usage of `sha256`, and its integration into public community projects](https://blog.syscall.party/2023/03/03/introducing-exphash-for-dll-clustering.html).

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -5900,7 +5900,7 @@ class PE:
         if len(export_list) == 0:
             return ""
 
-        return sha256(",".join(export_list).encode()).hexdigest()
+        return md5(",".join(export_list).encode()).hexdigest()
 
     def parse_import_directory(self, rva, size, dllnames_only=False):
         """Walk and parse the import directory."""


erocarrera:
I'm a bit torn in this case: I see the rationale in staying consistent with `imphash`, even more in the case that yara is also using `md5`.
Ideally all these methods would be using `sha256`, but one can't update `imphash` without causing major issues, I'd say it's not worth the pain.
Hence I'm leaning towards merging and using `md5` and hoping not many people have started relying on the `sha256` results.

PR #379, opened 2023-08-21
fix PE.get_data

mak:
Fix a situation when `length` passed to .get_data is 0 - expected results will be an empty  string but currently if rva is outside any section returned data will be unbounded


erocarrera:
Thanks!

PR #388, opened 2023-12-15
Fix ambiguous string syntax for PEid parsing regexp

hillu:
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1058741


erocarrera:
Thanks!

PR #427, opened 2024-09-11
Increase readability

j-t-1:
Make adjust_PointerToRawData and adjust_SectionAlignment more readable and consistent with each other. Also update date to 2024 in peutils.py.


j-t-1:
@erocarrera this commit breaks the pefile tests. Where are the tests for this located? Any help appreciated, as am unsure correct this.

PR #433, opened 2024-10-08
Simplify main

j-t-1:


nightlark:
Yea, / and * in function signatures may be useful. We might need to be careful with using those in existing functions to avoid breaking API changes. I could see them being used for new functions and if we added new optional arguments to existing functions.

PR #439, opened 2024-11-24
Tweak comments of generate_checksum

j-t-1:


pefile.py
```diff
@@ -7800,11 +7800,10 @@ def generate_checksum(self):
                 checksum = (checksum & 0xFFFFFFFF) + (checksum >> 32)
 
         checksum = (checksum & 0xFFFF) + (checksum >> 16)
-        checksum = (checksum) + (checksum >> 16)
-        checksum = checksum & 0xFFFF
+        checksum += (checksum >> 16)
+        checksum &= 0xFFFF
```
nightlark: Tweaking more than just comments here, but change seems reasonable and functionally equivalent.

j-t-1: Amended description.

PR #440, opened 2024-11-24
Small refactor of is_valid_function_name

j-t-1:
Slightly increases readability.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -2318,11 +2318,11 @@ def is_valid_function_name(
 ) -> bool:
     allowed_extra = b"._?@$()<>"
     if relax_allowed_characters:
-        allowed_extra = b"!\"#$%&'()*+,-./:<>?[\\]^_`{|}~@"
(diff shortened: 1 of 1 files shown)


pefile.py
```diff
@@ -2316,13 +2316,13 @@ def is_valid_dos_filename(s):
 def is_valid_function_name(
     s: Union[str, bytes, bytearray], relax_allowed_characters: bool = False
 ) -> bool:
-    allowed_extra = b"._?@$()<>"
+    allowed_extra = b"$().<>?@_"
     if relax_allowed_characters:
-        allowed_extra = b"!\"#$%&'()*+,-./:<>?[\\]^_`{|}~@"
```
nightlark: I'd be curious to know where these lists of characters originally came from. I'm not seeing many good references for the allowed set of characters in mangled function names, other than looking at the name mangling functions implemented in compilers and figuring out what set of characters can appear in them based on their code.

PR #443, opened 2024-11-27
Remove mode paramter

j-t-1:
The mode parameter of the open function defaults to 'r', open for reading, so does not need to be specified.


j-t-1:
> The flip side of this one is that being explicit can make it clearer what the intent was when opening the file.

Agree. Closed for the above reason.

PR #446, opened 2024-12-05
Format using f-strings

j-t-1:


pefile.py
```diff
@@ -2978,9 +2978,8 @@ def __unpack_data__(self, format, data, file_offset):
             structure.__unpack__(data)
         except PEFormatError as err:
             self.__warnings.append(
-                'Corrupt header "{0}" at file offset {1}. Exception: {2}'.format(
-                    format[0], file_offset, err
-                )
+                f'Corrupt header "{format[0]}" at file offset {file_offset}. '
+                f'Exception: {err}'.format
```
nightlark: Since a format string is being used, shouldn't the call to `.format` here be removed?

PR #456, opened 2024-12-23
Tweak peutils.py

j-t-1:
Small style changes.

### DIFF
--- a/peutils.py
+++ b/peutils.py
@@ -59,11 +59,11 @@ class SignatureDatabase:
         # - A dictionary with a string as a key (packer name)
         #   and None as value to indicate a full signature
         #
-        self.signature_tree_eponly_true = dict()
+        self.signature_tree_eponly_true = {}
         self.signature_count_eponly_true = 0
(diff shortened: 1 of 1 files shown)


peutils.py
```diff
@@ -582,7 +581,8 @@ def is_probably_packed(pe):
         if s_entropy > 7.4:
             total_compressed_data += s_length
 
-    if (total_compressed_data / total_pe_data_length) > 0.2:
-        has_significant_amount_of_compressed_data = True
+    has_significant_amount_of_compressed_data = (
+        total_compressed_data / total_pe_data_length > 0.2
+    )
```
nightlark: This feels like something where it could make sense to have the function take in optional threshold parameters instead of hard-coding 7.4 and 0.2 (that are customizable by the user). Perhaps something to consider for a future change.

nightlark: This PR looks like it can be rebased on master to resolve the merge conflict and pull in the changes from #471

Issue #396, opened 2024-03-18
256 MiB maximum offset limit is too low


randomascii:
The official dcheck enabled builds for Chrome are huge and have maximum section offsets of around 320 MB which is beyond the arbitrary hard-coded limit of 256 MiB in pefile.py. This causes symbol uploads to fail for these builds.

The limit should be increased to avoid this, and arguably the handling of the maximum should be changed so that an exception is thrown because the warning messages that are printed when invoking pefile.py are suppressed or otherwise hidden when using pefile.py as an import.

I have created a local change for Chrome that centralizes and increases the limit to 1 GiB.

randomascii:
The change I made in Chromium's copy of pefile.py can be found here:

https://chromium-review.googlesource.com/c/chromium/src/+/5377931

It's pretty simple - I just centralized the maximum offset limit and then raised it. I could easily make a PR for this change for here, if that would be accepted.

Or, maybe the limit should just be removed - maybe it has served its purpose and no longer makes sense.

Or, maybe the limit should be a constructor parameter. I haven't checked to see how much work this would be but at a high level this would be a cleaner solution.

Finally, if the limit is reached then I think an exception should be thrown so that pefile.py doesn't fail in silent and mysterious ways.

Any thoughts on this before I create a PR?

bluca:
I have opened a PR that adds a new parameter to let users configure the offset limit: https://github.com/erocarrera/pefile/pull/460

nightlark:
The PR that adds this as a configurable limit has been merged, will be included in the next release.

PR #460, opened 2025-01-30
Add max_offset parameter to configure hardcoded offset limit

bluca:
When parsing large images warnings are raised if they go beyond an hardcoded and fixed value. This is often not enough for many use cases, such as parsing a large Unified Kernel Image.

Add a parameter, defaulting to the previous value.

Fixes https://github.com/erocarrera/pefile/issues/396

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -2350,6 +2350,10 @@ class PE:
     whole PE structure. The "full_load" method can be used to parse
     the missing data at a later stage.
 
+    Warnings will be raised during parsing if a section is larger and/or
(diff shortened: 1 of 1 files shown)


pefile.py
```diff
@@ -3544,7 +3576,7 @@ def write(self, filename=None):
         with open(filename, "wb+") as f:
             f.write(new_file_data)
 
-    def parse_sections(self, offset):
+    def parse_sections(self, offset, max_offset=0x10000000):
```
j-t-1: `offset` is not defaulted here, so should `max_offset` be?

Issue #461, opened 2025-02-18
Adding `userforsecurity` var in `get_imphash(self)` function in `pefile.py` to bypass FIPS compliance


wangmot:
We are proposing to add a new `userforsecurity` variable in the `get_imphash(self)` function in `pefile.py` to allow bypassing FIPS compliance when calculating the imphash of a file. The pefile module currently uses the MD5 algorithm to calculate the imphash, but MD5 is banned in FIPS-compliant environments, which prevents pefile from running in those systems.

The `userforsecurity` flag would be particularly useful in environments where FIPS is enabled, as MD5 is not allowed. After Python 3.9, the `userforsecurity` variable can be set to False to bypass the FIPS restriction. We propose modifying the `get_imphash()` function to accept an optional `userforsecurity` parameter, which would allow bypassing FIPS compliance when set to False (with the default remaining True). This change would ensure that the `get_imphash()` function can be used in both FIPS-compliant and non-compliant environments without issues, providing the flexibility to disable FIPS compliance for MD5 when necessary. This solution would address the current limitation and improve compatibility with systems that require MD5 hash functionality.

PR #462, opened 2025-02-18
add `usedforsecurity` flag to bypass FIPS machines

wangmot:
`get_imphash()` in `pefile.py` uses md5 algorithm to calculate the imphash of a file. FIPS compliant machines don't work if md5 hash is in use so the pefile package doesn't run in those environments. Having the `usedforsecurity` flag set to False will bypass this and allow it to run on FIPS machines.  

Please see related issue here: https://github.com/erocarrera/pefile/issues/461

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -13,11 +13,11 @@ PEs as well as malware, which often attempts to abuse the format way beyond its
 standard use. To the best of my knowledge most of the abuse is handled
 gracefully.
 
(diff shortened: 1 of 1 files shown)


j-t-1:
`get_imphash(self, usedforsecurity=True): # Default for usedforsecurity set to True`

Could remove the comment, somewhat self-explanatory?

wangmot:
@erocarrera Hey, anything else you need for this PR?

PR #466, opened 2025-08-04
Continue instead of break out of loop in parse_relocations_with_format

cap2k4-rivos:
In RISC-V there can be duplicate relocation entries, we want to continue parsing instead of breaking in this case.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -4353,7 +4353,7 @@ class PE:
                     "Overlapping offsets in relocation data "
                     "at RVA: 0x%x" % (reloc_offset + rva)
                 )
-                break
+                continue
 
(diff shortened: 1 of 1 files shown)


j-t-1:
This seems to be because of instructions like auipc, and this interacts with relocations.

> RISC-V leverages the linker's relocation infrastructure more aggressively than any other architecture

https://www.sifive.com/blog/all-aboard-part-2-relocations

Dynamic relocations are done by the loader.

We could have a comment within the code about relocations in RISC-V, comparing it to other architectures.

@cap2k4-rivos do you have a link about this, potentially to put as a comment in the code to document the change?

PR #493, opened 2026-03-31
Improve finding the Rich Header data

### LINKED ISSUE #272: Rich Header offset is NOT static
Currently, `pefile` only expects a rich header at offset `0x80`:
https://github.com/erocarrera/pefile/blob/bbf28920a71248ed5c656c81e119779c131d9bd4/pefile.py#L2357

But the rich header is not always at `0x80`. Here are a couple examples, both malicious and benign, of this:
- https://www.virustotal.com/gui/file/582ce3eea9c97d5e89f7d83953a6d518b16770e635a19a456c0225449c6967a4/content/hex
- https://www.virustotal.com/gui/file/eb5fea95a9e8bb05dd0f111d35fa97bd3b75aba07a5b40e5ba7eae97a156e373/content/hex

In both of these examples, the header begins at offset `0x200`. It is unclear whether `0x80` and `0x200` are the only two possibilities, or if more exist. Regardless, it seems like the issue could be fixed if we strip leading null bytes from the `rich_data`: 
```python
rich_data = self.get_data(0x80, rich_index + 8).lstrip(b'\x00')
```

My only concern would be if it's possible for a rich header to begin with a null byte, but I'm not sure on that.

j-t-1:
See Issue 272.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -3394,6 +3394,7 @@ class PE:
     def parse_rich_header(self):
         """Parses the Rich Header
         https://www.ntcore.com/files/richsign.htm
+        https://www.virusbulletin.com/virusbulletin/2020/01/vb2019-paper-rich-headers-leveraging-mysterious-artifact-pe-format
 
         Structure:
         00 DanS ^ checksum, checksum, checksum, checksum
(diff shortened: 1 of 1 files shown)


j-t-1:
@nightlark we only test `parse_rich_header` on kernel32.dll (via `get_rich_header_hash`).

All the test files that have a Rich Header (90) are at offset 0x80. I tried C:\Windows\System32\drivers\beep.sys but this also is at offset 0x80.

This PR was checked: https://github.com/erocarrera/pefile/issues/272#issuecomment-4216254098.

PR #498, opened 2026-04-09
Rename functions names and use an f-string

j-t-1:


nightlark (approved):
I’m kinda ambivalent about the function name changes. Technically it would be a breaking change for anyone using it, but it also isn’t clear that it is used by people and is primarily just internal usage by pefile.

If we’re going to change the names it probably does make sense to do it now along with other potential breaking changes in the upcoming release.

j-t-1:
> If we’re going to change the names it probably does make sense to do it now along with other potential breaking changes in the upcoming release.

Was thinking the same. Afterwards the interfaces should be more stable.

j-t-1:
The `format_ordinal` function adds noise to what is just a string encode. Will probably remove it when doing the further update.

nightlark:
format_ordinal is also adding the “ord” prefix

j-t-1:
Yes, so not quite as obvious change as a string encode. Still may be better replacing it as the function call is about the same size as doing the conversion in place.

PR #500, opened 2026-04-09
Freeze ordinal to name mapping used for imphash calculations

### LINKED ISSUE #141: Imphash implementation does not follows convention
https://www.mandiant.com/blog/tracking-malware-import-hashing/

> Mandiant's imphash convention requires the following:
   * Resolving ordinals to function names when they appear
   * Converting both DLL names and function names to all lowercase
   * Removing the **file extensions** from imported module names
   * Building and storing the lowercased strings in an ordered list
   * Generating the MD5 hash of the ordered list

pefile only removes the following extensions: [ocx, sys, dll](https://github.com/erocarrera/pefile/blob/master/pefile.py#L3666-L3676). Any reason to do that? What about **exe**, **drv**, etc? Why not to remove any extension?

nightlark:
This makes a frozen-in-time copy of the ordinal to name mappings to use when calculating the imphash, and ensures that any function names from newer ordinal name mappings don't get used when calculating the imphash.

Resolves #141


pefile.py
```diff
@@ -5822,8 +5823,15 @@ def get_imphash(self, usedforsecurity=True):
             entry_dll_lower = entry.dll.lower()
             for imp in entry.imports:
                 funcname = None
-                if not imp.name:
-                    funcname = ordlookup.ordinal_lookup(
+                if not imp.name or getattr(imp, "name_from_ordinal", False):
+                    # Must use imphash-specific ordinal lookup whose tables are
+                    # frozen to the state when the imphash algorithm was first
+                    # implemented in pefile, so that imphash values remain compatible
+                    # with YARA, YARA-X, and other tools based on the original pefile
```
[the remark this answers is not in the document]
nightlark: Removing only ocx, sys, and dll extensions matches the imphash implementation in other tools (YARA, YARA-X) - at this point no changes can be made to that behavior without affecting the values calculated.

ordlookup/__init__.py
```diff
@@ -29,3 +29,34 @@ def ordinal_lookup(libname, ord_val, make_name=False):
...
+    implementation.
+    """
+    names = ords.get(libname.lower())
+    if names is None:
+        if make_name is True:
+            return f"ord{ord_val}".encode()
+        return None
+    name = names.get(ord_val)
+    if name is None:
+        return f"ord{ord_val}".encode()
```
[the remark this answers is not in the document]
nightlark: Fixed the issue with not using `imphash_ords`. Could go either way with using `format_ordinal` -- though I'd probably want to use a new `imphash_format_ordinal` function to ensure it never gets updated.

ordlookup/__init__.py
```diff
@@ -29,3 +29,34 @@ def ordinal_lookup(libname, ord_val, make_name=False):
     if name is None:
         return format_ordinal(ord_val)
     return name
+
+
+# Imphash-specific ordinal tables must remain frozen to the state when the original
+# implementation was written so that imphash values remain compatible with YARA,
+# YARA-X, and other tools based on the original pefile implementation.  wsock32 is
```
j-t-1, line 36:
```suggestion
# YARA-X, and other tools based on the original pefile implementation. wsock32 is
```

ordlookup/imphash_oleaut32.py
```diff
@@ -0,0 +1,405 @@
+# DO NOT MODIFY THE ORDINAL MAPPINGS IN THIS FILE
```
j-t-1, line 1: Maybe insert a new line here.

ordlookup/imphash_oleaut32.py
```diff
@@ -0,0 +1,405 @@
...
+# These are used for calculating imphashes, and changing them will change the imphash values for all files
+# that use this module.
```
j-t-1, line 3: Maybe insert a new line here.

j-t-1 (approved):
The new lines is a suggestion, applicable to the two new files. This is marginal, keep as-is if you prefer.

nightlark:
@wxsBSD and @plusvic - would welcome some more eyes on this to double/triple check that the imphash calculation produces values that match the original implementation/YARA.

wxsBSD:
I think this is good, and restores the original behavior.

nightlark:
I made the suggested changes, will merge once tests finish running (if they pass).

PR #502, opened 2026-04-10
Set up Dependabot for keeping workflow and pip dependencies updated

nightlark:
Sets up Dependabot for keeping GitHub actions workflow dependencies, and Python project dependencies up to date. Resolves https://github.com/erocarrera/pefile/security/code-scanning/2 code-scanning alert.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
j-t-1, line 11: Because we do not have many dependencies, weekly seems a good interval.

j-t-1 (approved):

.github/dependabot.yml
nightlark: Yea, for users there aren’t any. Just a few dev ones.

PR #522, opened 2026-04-22
Broaden definition of a pageable section

j-t-1:
A driver developer has the option to make designated parts of a driver
pageable so that Windows can move these parts to the paging file when
they are not in use. To make a code or data section pageable, the driver
developer assigns a name that begins with "PAGE" to the section. The I/O
manager checks the names of the sections when it loads a driver. If a
section name begins with "PAGE", the I/O manager makes the section
pageable.
https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/making-drivers-pageable

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -7886,9 +7886,8 @@ class PE:
         ):
             return True
 
-        driver_like_section_names = {b"page", b"paged"}
-        if driver_like_section_names.intersection(
-            {section.Name.lower().rstrip(b"\x00") for section in self.sections}
+        if any(
+            {section.Name.lower().startswith(b"page") for section in self.sections}
         ) and (
             self.OPTIONAL_HEADER.Subsystem
             in (


nightlark:
Do we have test files for drivers that do this??

j-t-1:
> Do we have test files for drivers that do this??

None trigger the if statement before or after this change, although I do not know how many are drivers (there are only four test files with a sys extension).

I tried beep.sys, which I think is pageable, but this does not have section names starting with "PAGE".

From Microsoft documentation on __declspec(code_seg):

> The code_seg declaration attribute names an executable text segment in the .obj file in which the object code for the function or class member functions is stored.
> 
> Syntax
> `__declspec(code_seg("segname")) declarator`

Their example uses "PagedMem":

`auto Sqr = [](int t) __declspec(code_seg("PagedMem")) -> int { return t*t; };`

The existing code replaced in this PR would not classify it as a driver. (Although the linker does not necessarily preserve names.)

PR #533, opened 2026-05-31
Modify functools.lru_cache use

j-t-1:
Instead of amending functools.lru_cache, create a new function, lru_cache_copy. This makes the implementation simpler and is more obvious when it is used.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -21,8 +21,7 @@ __version__ = "2024.8.26"
 __contact__ = "ero.carrera@gmail.com"
 
 import codecs
-import copy as copymod
-import functools
+import copy
 import math
 import mmap
 import os
@@ -31,6 +30,7 @@ import struct
 import time
 import uuid
 from collections import Counter, defaultdict
+from functools import lru_cache, wraps
 from hashlib import md5, sha1, sha256, sha512
 
 import ordlookup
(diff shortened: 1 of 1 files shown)


pefile.py
```diff
@@ -31,29 +30,25 @@
...
-    if not copy:
-        return functools.lru_cache(maxsize, typed)
-
+def lru_cache_copy(maxsize=128, typed=False):
     def decorator(f):
-        cached_func = functools.lru_cache(maxsize, typed)(f)
+        cached_function = lru_cache(maxsize, typed)(f)
 
-        @functools.wraps(f)
+        @wraps(f)
```
nightlark, line 49: wraps as a name feels on the generic side, maybe change it to functools_wraps:

```python
from functools import lru_cache, wraps as functools_wraps
```

pefile.py
```diff
@@ -21,8 +21,7 @@
 __contact__ = "ero.carrera@gmail.com"
 
 import codecs
-import copy as copymod
```
nightlark, line 24: Importing with the changed name might actually be to avoid potential confusion between the copy module in case a future change creates a variable named copy.

j-t-1: Yes, removing the copy variable allowed it to be imported as is, and a future change may need it to be renamed. Can be done in the future PR as needed to keep together.

pefile.py
j-t-1: Yes, this is a slight disadvantage here. Calling the two imported functions differently: `lru_cache` and `functools.wraps` may be useful, with the imports:
```
import functools
from functools import lru_cache
```

Conversely, since they are both decorators and from the same module, importing them and using them as is gives consistency.

Will keep as is, but open to it be changed in another PR.

PR #542, opened 2026-06-25
fix: respect VirtualSize when building the memory-mapped image (#392)

### LINKED ISSUE #392: get_memory_mapped_image leaves trash data in alignment regions
In a memory mapped image, regions that exist due to section/page alignment requirements should contain zeroes. Currently, pefile bleeds section data into such regions.

Reproduction: Take any Windows binary that has file-aligned sections (so not a memory dump) and check what's at offset 0x400 (which in most common binaries is where the PE header ends).
```python
>>> import pefile
>>> pe = pefile.PE(name="C:\\Windows\\explorer.exe")
>>> image = pe.get_memory_mapped_image()
>>> assert image[0x400:0x1000] != image[0x1000:0x1C00]
AssertionError
```

This can cause considerable confusion if you're doing a regex search over the mapped image and RIP-relative data references get screwed up.

gaoflow:
## Summary

Fixes #392.

`get_memory_mapped_image()` returned raw section data up to `SizeOfRawData`
bytes for every section, ignoring `Misc_VirtualSize`. The PE/COFF spec
(and the Windows loader) define two distinct behaviors:

| Condition | Expected mapped content |
|---|---|
| `VirtualSize < SizeOfRawData` | Only `VirtualSize` bytes are mapped; the remaining disk bytes are file-alignment padding and must **not** appear in the memory view |
| `VirtualSize > SizeOfRawData` | `SizeOfRawData` bytes are mapped, then the section is **zero-padded** to `VirtualSize` (BSS region) |

### Root cause

The single line responsible:

```python
mapped_data += section.get_data()   # always returns SizeOfRawData bytes
```

`section.get_data()` (without `ignore_padding=True`) returns exactly `SizeOfRawData`
bytes, so file-alignment padding silently bleeds into the mapped view whenever
`VirtualSize < SizeOfRawData`, and the BSS tail is truncated whenever
`VirtualSize > SizeOfRawData`.

### Fix

After fetching the raw section data, clip it to `VirtualSize` (drops disk
padding) and zero-pad it to `VirtualSize` if shorter (fills BSS):

```python
section_data = section.get_data()
if section.Misc_VirtualSize:
    if len(section_data) > section.Misc_VirtualSize:
        section_data = section_data[: section.Misc_VirtualSize]
    elif len(section_data) < section.Misc_VirtualSize:
        section_data += b"\0" * (section.Misc_VirtualSize - len(section_data))
mapped_data += section_data
```

Two regression tests covering both directions of the inequality are added to
`tests/export_test.py` using a synthetic, self-contained PE binary so they
require no external test files.

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -6325,7 +6325,19 @@ class PE:
             elif padding_length < 0:
                 mapped_data = mapped_data[:padding_length]
 
-            mapped_data += section.get_data()
+            section_data = section.get_data()
+            virtual_size = section.Misc_VirtualSize
+            if virtual_size and len(section_data) != virtual_size:
+                # Only Misc_VirtualSize bytes belong in the mapped view.
+                if len(section_data) > virtual_size:
+                    # VirtualSize < SizeOfRawData: the extra disk bytes are
+                    # file-alignment padding the OS loader discards.
+                    section_data = section_data[:virtual_size]
+                else:
+                    # VirtualSize > SizeOfRawData: the gap is BSS and must be
+                    # zero-padded in the mapped view.
+                    section_data += b"\0" * (virtual_size - len(section_data))
+            mapped_data += section_data
 
         # If the image was rebased, restore it to its original form
         if ImageBase is not None:
--- a/pefile.py
+++ b/pefile.py
@@ -6329,14 +6329,14 @@ class PE:
             virtual_size = section.Misc_VirtualSize
             if virtual_size and len(section_data) != virtual_size:
                 # Only Misc_VirtualSize bytes belong in the mapped view.
-                if len(section_data) > virtual_size:
+                if virtual_size < len(section_data):
                     # VirtualSize < SizeOfRawData: the extra disk bytes are
                     # file-alignment padding the OS loader discards.
                     section_data = section_data[:virtual_size]
-                else:
+                elif virtual_size > len(section_data):
                     # VirtualSize > SizeOfRawData: the gap is BSS and must be
                     # zero-padded in the mapped view.
-                    section_data += b"\0" * (virtual_size - len(section_data))
+                    section_data += b"\x00" * (virtual_size - len(section_data))
             mapped_data += section_data
 
         # If the image was rebased, restore it to its original form
--- a/tests/pefile_test.py
+++ b/tests/pefile_test.py
@@ -776,7 +776,7 @@ def _create_pe(virtual_size, size_of_raw_data):
     headers = headers.ljust(pointer_to_raw_data, b"\x00")
 
     # Fill raw section data with a sentinel byte so leakage is obvious
-    section_data = b"\xcc" * size_of_raw_data
+    section_data = b"\xCC" * size_of_raw_data
     return bytes(headers) + section_data
 
 
@@ -797,7 +797,7 @@ class Test_memory_mapped_image(unittest.TestCase):
         # Section content up to VirtualSize must be present.
         self.assertEqual(
             image[va : va + vsize],
-            b"\xcc" * vsize,
+            b"\xCC" * vsize,
             "section content must be preserved up to VirtualSize",
         )
         # The image must not extend past VirtualAddress + VirtualSize; raw
@@ -811,7 +811,7 @@ class Test_memory_mapped_image(unittest.TestCase):
     def test_virtual_size_greater_than_raw_size(self):
         """Uninitialized BSS region must be zero-padded in the memory-mapped image.
 
-        When VirtualSize > SizeOfRawData the bytes from SizeOfRawData up to
+        When SizeOfRawData < VirtualSize the bytes from SizeOfRawData up to
         VirtualSize represent BSS and must be zero in the mapped view.
         """
         vsize = 0x1500
@@ -822,7 +822,7 @@ class Test_memory_mapped_image(unittest.TestCase):
         va = pe.sections[0].VirtualAddress
         self.assertEqual(
             image[va : va + raw_size],
-            b"\xcc" * raw_size,
+            b"\xCC" * raw_size,
             "raw section content must be preserved",
         )
         self.assertEqual(
--- a/tests/pefile_test.py
+++ b/tests/pefile_test.py
@@ -344,6 +344,56 @@ class TestPEFile(unittest.TestCase):
         control_file = os.path.join(REGRESSION_TESTS_DIR, "empty_file")
         self.assertRaises(pefile.PEFormatError, pefile.PE, control_file)
 
+    def test_virtual_size_less_than_raw_size(self):
+        """File-alignment padding must not bleed into the memory-mapped image.
+
+        When VirtualSize < SizeOfRawData the bytes beyond VirtualSize are
+        disk padding that the OS loader never maps.  The memory-mapped image
+        must end at VirtualAddress + VirtualSize, not VirtualAddress + SizeOfRawData.
+        """
+        vsize = 0x800
+        raw_size = 0x1000
+        pe = pefile.PE(data=_create_pe(vsize, raw_size))
+        image = pe.get_memory_mapped_image()
+
+        va = pe.sections[0].VirtualAddress
+        # Section content up to VirtualSize must be present.
+        self.assertEqual(
+            image[va : va + vsize],
+            b"\xCC" * vsize,
+            "section content must be preserved up to VirtualSize",
+        )
+        # The image must not extend past VirtualAddress + VirtualSize; raw
+        # file-padding bytes (0xCC beyond the boundary) must not be included.
+        self.assertEqual(
+            len(image),
+            va + vsize,
+            "mapped image must not include raw file-padding past VirtualSize",
+        )
+
+    def test_virtual_size_greater_than_raw_size(self):
+        """Uninitialized BSS region must be zero-padded in the memory-mapped image.
+
+        When SizeOfRawData < VirtualSize the bytes from SizeOfRawData up to
+        VirtualSize represent BSS and must be zero in the mapped view.
+        """
+        vsize = 0x1500
+        raw_size = 0x1000
+        pe = pefile.PE(data=_create_pe(vsize, raw_size))
+        image = pe.get_memory_mapped_image()
+
+        va = pe.sections[0].VirtualAddress
+        self.assertEqual(
+            image[va : va + raw_size],
+            b"\xCC" * raw_size,
+            "raw section content must be preserved",
+        )
+        self.assertEqual(
+            image[va + raw_size : va + vsize],
+            b"\x00" * (vsize - raw_size),
+            "BSS region (VirtualSize - SizeOfRawData) must be zero-padded",
+        )
+
     def test_relocated_memory_mapped_image(self):
         """Test different rebasing methods produce the same image"""
 
@@ -749,7 +799,7 @@ def _create_pe(virtual_size, size_of_raw_data):
         + struct.pack("<HHHHHH", 6, 0, 0, 0, 6, 0)  # OS/Image/Subsystem versions
         + struct.pack("<I", 0)  # Win32VersionValue
         + struct.pack("<I", size_of_image)  # SizeOfImage
-        + struct.pack("<I", pointer_to_raw_data)  # SizeOfHeaders
+        + struct.pack("<I", 0x200)  # SizeOfHeaders (headers rounded up to FileAlignment)
         + struct.pack("<I", 0)  # CheckSum
         + struct.pack("<H", 2)  # Subsystem GUI
         + struct.pack("<H", 0x8100)  # DllCharacteristics
@@ -778,55 +828,3 @@ def _create_pe(virtual_size, size_of_raw_data):
     # Fill raw section data with a sentinel byte so leakage is obvious
     section_data = b"\xCC" * size_of_raw_data
     return bytes(headers) + section_data
-
-
-class Test_memory_mapped_image(unittest.TestCase):
-    def test_virtual_size_less_than_raw_size(self):
-        """File-alignment padding must not bleed into the memory-mapped image.
-
-        When VirtualSize < SizeOfRawData the bytes beyond VirtualSize are
-        disk padding that the OS loader never maps.  The memory-mapped image
-        must end at VirtualAddress + VirtualSize, not VirtualAddress + SizeOfRawData.
-        """
-        vsize = 0x800
-        raw_size = 0x1000
-        pe = pefile.PE(data=_create_pe(vsize, raw_size))
-        image = pe.get_memory_mapped_image()
-
-        va = pe.sections[0].VirtualAddress
-        # Section content up to VirtualSize must be present.
-        self.assertEqual(
-            image[va : va + vsize],
-            b"\xCC" * vsize,
-            "section content must be preserved up to VirtualSize",
-        )
-        # The image must not extend past VirtualAddress + VirtualSize; raw
-        # file-padding bytes (0xCC beyond the boundary) must not be included.
-        self.assertEqual(
-            len(image),
-            va + vsize,
-            "mapped image must not include raw file-padding past VirtualSize",
-        )
-
-    def test_virtual_size_greater_than_raw_size(self):
-        """Uninitialized BSS region must be zero-padded in the memory-mapped image.
-
-        When SizeOfRawData < VirtualSize the bytes from SizeOfRawData up to
-        VirtualSize represent BSS and must be zero in the mapped view.
-        """
-        vsize = 0x1500
-        raw_size = 0x1000
-        pe = pefile.PE(data=_create_pe(vsize, raw_size))
-        image = pe.get_memory_mapped_image()
-
-        va = pe.sections[0].VirtualAddress
-        self.assertEqual(
-            image[va : va + raw_size],
-            b"\xCC" * raw_size,
-            "raw section content must be preserved",
-        )
-        self.assertEqual(
-            image[va + raw_size : va + vsize],
-            b"\x00" * (vsize - raw_size),
-            "BSS region (VirtualSize - SizeOfRawData) must be zero-padded",
-        )
--- a/tests/pefile_test.py
+++ b/tests/pefile_test.py
@@ -4,6 +4,7 @@ import struct
 import sys
 import unittest
 from hashlib import sha256
+import struct
 
 import pefile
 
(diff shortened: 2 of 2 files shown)


j-t-1:
Thanks @gaoflow.

Reading this I was thinking `section.get_data()` could be updated so it does this, but I think the way you have done may be simpler and clearer.

Above you mention BSS, which would be the obvious example section where `VirtualSize > SizeOfRawData`, so although the fix works for any section that satisfies this inequality mentioning BSS clarifies when this would be mainly be used.

tests/export_test.py
```diff
@@ -115,6 +116,120 @@ def test_exports64(self):
...
+        + b"\x00" * 128  # DataDirectory (16 entries, all zero)
+    )
+    assert len(opt_header) == 0xE0
+
+    section_header = (
+        b".text\x00\x00\x00"  # Name
+        + struct.pack("<I", virtual_size)
+        + struct.pack("<I", va)
+        + struct.pack("<I", size_of_raw_data)
+        + struct.pack("<I", raw_ptr)
```
j-t-1: Does raw_ptr equals PointerToRawData? Rename to pointer_to_raw_data for readability.

tests/export_test.py
```diff
@@ -115,6 +116,120 @@ def test_exports64(self):
...
+        + struct.pack("<IHHHI", 0, 0, 0, 0, 0x60000020)
```
j-t-1: The format string should be "<IIHHI"?

pefile.py
```diff
@@ -6323,7 +6323,17 @@ def get_memory_mapped_image(self, max_virtual_address=0x10000000, ImageBase=None
...
                 mapped_data = mapped_data[:padding_length]
 
-            mapped_data += section.get_data()
+            section_data = section.get_data()
+            # In a memory-mapped image only Misc_VirtualSize bytes of each section
+            # are visible; raw data beyond that boundary is file-alignment padding
+            # that the OS loader discards.  If VirtualSize < SizeOfRawData, trim the
+            # extra bytes.  If VirtualSize > SizeOfRawData, zero-pad up to VirtualSize.
+            if section.Misc_VirtualSize:
+                if len(section_data) > section.Misc_VirtualSize:
```
j-t-1: Switch inequality to match comment.

pefile.py
```diff
@@ -6323,7 +6323,17 @@ def get_memory_mapped_image(self, max_virtual_address=0x10000000, ImageBase=None
...
+                    section_data = section_data[: section.Misc_VirtualSize]
+                elif len(section_data) < section.Misc_VirtualSize:
```
j-t-1: Switch inequality to match comment.

j-t-1 (changes requested):
Like the approach.

tests/export_test.py
```diff
@@ -1,3 +1,4 @@
+import struct
```
j-t-1: Move the tests to pefile_test.py, probably under `test_relocated_memory_mapped_image`.

tests/export_test.py
```diff
@@ -115,6 +116,120 @@ def test_exports64(self):
         self.assertEqual(EXPECT_NAMES, exports)
 
 
+def _make_pe_with_virtual_size(virtual_size, size_of_raw_data):
```
j-t-1: Add this to the end of pefile_test.py, with a rename to `_create_pe` as we may be able to reuse it for other existing or new tests.

j-t-1:
@nightlark the test @gaoflow has done uses a function to create a non-functional PE file. What do you think of making it more generic and thus able to be used for other tests? Outside of this PR (but maybe just changing this PR so it the arguments are keyword only).

binggao1230:
Thanks @j-t-1 — all addressed in `e4b5ed2`:

- **Inequalities in `pefile.py`**: rewrote the guard so `section.Misc_VirtualSize` is bound once to `virtual_size` and the comparisons read `len > virtual_size` / `len < virtual_size`, matching the comment.
- **Format string**: fixed to `<IIHHI` — the section-header tail is `PointerToRelocations(I) PointerToLinenumbers(I) NumberOfRelocations(H) NumberOfLinenumbers(H) Characteristics(I)`. The earlier `<IHHHI` mislaid the `NumberOfLinenumbers` half-word.
- **`raw_ptr` naming**: renamed to `pointer_to_raw_data` throughout the builder.
- **Tests moved**: the two new regression tests now live in `pefile_test.py` (`Test_memory_mapped_image`), alongside `test_relocated_memory_mapped_image`; the helper is renamed to `_create_pe` so it can be reused by other tests. `export_test.py` is back to upstream.

Both new tests pass locally (`python -m unittest pefile_test.Test_memory_mapped_image`). Happy to adjust anything else.

This pull request was prepared with the assistance of AI, under my direction and review.

nightlark:
> @nightlark the test @gaoflow has done uses a function to create a non-functional PE file. What do you think of making it more generic and thus able to be used for other tests? Outside of this PR (but maybe just changing this PR so it the arguments are keyword only).

Yea, this is an interesting approach that could give some tests tailored to checking specific cases without resorting to the regression tests.

> Above you mention BSS, which would be the obvious example section where `VirtualSize > SizeOfRawData`, so although the fix works for any section that satisfies this inequality mentioning BSS clarifies when this would be mainly be used.

I'm reading up on what bytes should be file-backed vs zero-filled, and there's some evidence that the Windows loader rounds up to the page size, which would require a cap more along the lines of `min(SizeOfRawData, roundup(VirtualSize, SectionAlignment))` -- for "correct" binaries the decision might not matter, but there are some corkami binaries that intentionally use the space to hide data.

Still trying to wrap my head around it all, but we want to avoid making a change that will overcorrect and hide bytes that should be visible. (This is an interesting adventure that has led me to the Windows Research Kernel; working on getting a test binary that I can open up and inspect in a debugger on Windows to verify that I understand what is going on).

> Reading this I was thinking `section.get_data()` could be updated so it does this, but I think the way you have done may be simpler and clearer.

We might want to be careful of pre-initializing the data up to VirtualSize, this could result in a large data allocation when looking at binaries that intentionally have a bad VirtualSize.

tests/pefile_test.py
```diff
@@ -658,3 +659,124 @@ def test_checksum(self):
...
+        """
+        vsize = 0x1500
+        raw_size = 0x1000
+        pe = pefile.PE(data=_create_pe(vsize, raw_size))
+        image = pe.get_memory_mapped_image()
+
+        va = pe.sections[0].VirtualAddress
+        self.assertEqual(
+            image[va : va + raw_size],
+            b"\xcc" * raw_size,
```
j-t-1:
```suggestion
            b"\xCC" * raw_size,
```

pefile.py
```diff
@@ -6323,7 +6323,19 @@ def get_memory_mapped_image(self, max_virtual_address=0x10000000, ImageBase=None
...
+            if virtual_size and len(section_data) != virtual_size:
+                # Only Misc_VirtualSize bytes belong in the mapped view.
+                if len(section_data) > virtual_size:
+                    # VirtualSize < SizeOfRawData: the extra disk bytes are
+                    # file-alignment padding the OS loader discards.
+                    section_data = section_data[:virtual_size]
+                else:
+                    # VirtualSize > SizeOfRawData: the gap is BSS and must be
+                    # zero-padded in the mapped view.
+                    section_data += b"\0" * (virtual_size - len(section_data))
```
j-t-1:
```suggestion
                    section_data += b"\x00" * (virtual_size - len(section_data))
```

tests/pefile_test.py
```diff
@@ -658,3 +659,124 @@ def test_checksum(self):
...
+        vsize = 0x800
+        raw_size = 0x1000
+        pe = pefile.PE(data=_create_pe(vsize, raw_size))
+        image = pe.get_memory_mapped_image()
+
+        va = pe.sections[0].VirtualAddress
+        # Section content up to VirtualSize must be present.
+        self.assertEqual(
+            image[va : va + vsize],
+            b"\xcc" * vsize,
```
j-t-1:
```suggestion
            b"\xCC" * vsize,
```

tests/pefile_test.py
```diff
@@ -658,3 +659,124 @@ def test_checksum(self):
...
+        # NumberOfRelocations, NumberOfLinenumbers, Characteristics
+    )
+
+    headers = (
+        bytes(dos_header) + b"PE\x00\x00" + file_header + opt_header + section_header
+    )
+    headers = headers.ljust(pointer_to_raw_data, b"\x00")
+
+    # Fill raw section data with a sentinel byte so leakage is obvious
+    section_data = b"\xcc" * size_of_raw_data
```
j-t-1:
```suggestion
    section_data = b"\xCC" * size_of_raw_data
```

tests/pefile_test.py
```diff
@@ -658,3 +659,124 @@ def test_checksum(self):
...
+        self.assertEqual(
+            len(image),
+            va + vsize,
+            "mapped image must not include raw file-padding past VirtualSize",
+        )
+
+    def test_virtual_size_greater_than_raw_size(self):
+        """Uninitialized BSS region must be zero-padded in the memory-mapped image.
+
+        When VirtualSize > SizeOfRawData the bytes from SizeOfRawData up to
```
j-t-1:
```suggestion
        When SizeOfRawData < VirtualSize the bytes from SizeOfRawData up to
```

pefile.py
```diff
@@ -6323,7 +6323,19 @@ def get_memory_mapped_image(self, max_virtual_address=0x10000000, ImageBase=None
             elif padding_length < 0:
                 mapped_data = mapped_data[:padding_length]
 
-            mapped_data += section.get_data()
+            section_data = section.get_data()
+            virtual_size = section.Misc_VirtualSize
+            if virtual_size and len(section_data) != virtual_size:
+                # Only Misc_VirtualSize bytes belong in the mapped view.
+                if len(section_data) > virtual_size:
```
j-t-1:
```suggestion
                if virtual_size < len(section_data):
```

pefile.py
```diff
@@ -6323,7 +6323,19 @@ def get_memory_mapped_image(self, max_virtual_address=0x10000000, ImageBase=None
...
+                    # VirtualSize < SizeOfRawData: the extra disk bytes are
+                    # file-alignment padding the OS loader discards.
+                    section_data = section_data[:virtual_size]
+                else:
```
j-t-1:
```suggestion
                elseif virtual_size > len(section_data):
```

j-t-1 (commented):
Small non-functional suggestions.

Also move `_create_pe` to the end of pefile_test.py.

I think move `test_virtual_size_less_than_raw_size` and `test_virtual_size_greater_than_raw_size` between `test_empty_file_exception` and `test_relocated_memory_mapped_image`, so we just have one class (removing Test_memory_mapped_image). Unsure if having just one class is best practice, but keeps with existing way.

j-t-1:
> Yea, this is an interesting approach that could give some tests tailored to checking specific cases without resorting to the regression tests.

Agree. Thanks @gaoflow.

> I'm reading up on what bytes should be file-backed vs zero-filled, and there's some evidence that the Windows loader rounds up to the page size, which would require a cap more along the lines of min(SizeOfRawData, roundup(VirtualSize, SectionAlignment)) -- for "correct" binaries the decision might not matter, but there are some corkami binaries that intentionally use the space to hide data.

I did not think of this, you are likely correct that the memory mapped section is a multiple of page size.

binggao1230:
Thanks for the suggestions — applied in 243ab13:

- Reordered the mapped-image branch to read `if virtual_size < len(section_data)` / `elif virtual_size > len(section_data)`, so each condition matches the VirtualSize-first wording of its comment (behaviour is unchanged — the outer `len != virtual_size` guard already rules out equality, so the `elif` covers the same case the `else` did).
- Switched the zero-pad to an explicit `b"\x00"` and the test sentinel to uppercase `b"\xCC"`.
- Tweaked the BSS test docstring to "SizeOfRawData < VirtualSize".

The earlier round (test moved into `pefile_test.py` as `_create_pe`, `pointer_to_raw_data` rename, `<IIHHI` format) is already in the branch. `test_virtual_size_less_than_raw_size` and `test_virtual_size_greater_than_raw_size` both pass locally.

j-t-1:
Thanks @gaoflow.

Can now change this line:
`if virtual_size and len(section_data) != virtual_size:`
to
`if virtual_size:`
I do not know if can be removed completely, but safer to keep as is for now.


The program runs at the command line.
The INT3 instruction 0xCC is a one-byte-instruction used by debuggers, so could be an excellent choice. 

Replacing 0xCC with the below also ran, but 0xCC may be better so keep as is.

```
# Append section data with NOPs so leakage is obvious
section_data = b"\x90" * (size_of_raw_data - 6)  # NOPs
# Return 0x100 for minimal check the program ran
section_data += b"\xB8\x00\x01\x00\x00"  #  mov eax, 0x100
section_data += b"\xC3"  # ret
```

Then get the memory-mapped information when it is run (@nightlark is there an easy way?) so we can check the PR calculation.

nightlark:
What I'm thinking to check it is create a minimal binary that takes advantage of the extra space in a way that makes it obvious where the cut-off is for what parts of the memory mapping are backed by the file, and use x64dbg to inspect what's actually loaded into memory.

I think this section in the WRK is relevant for how Windows is handling the memory mapping: https://github.com/ayyucedemirbas/Windows-Research-Kernel-WRK-/blob/master/WRK-v1.2/base/ntos/mm/creasect.c#L3354-L3401

j-t-1:
> What I'm thinking to check it is create a minimal binary that takes advantage of the extra space in a way that makes it obvious where the cut-off is for what parts of the memory mapping are backed by the file, and use x64dbg to inspect what's actually loaded into memory.

The creator/maintainer of x64dbg has contributed to pefile, as you may have known.

> I think this section in the WRK is relevant for how Windows is handling the memory mapping: https://github.com/ayyucedemirbas/Windows-Research-Kernel-WRK-/blob/master/WRK-v1.2/base/ntos/mm/creasect.c#L3354-L3401

Nice, below is for understanding.

Those lines are within NtCreateSection, and this routine creates a section object.
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntcreatesection

A section object represents a section of memory that can be shared. A process can use a section object to share parts of its memory address space (memory sections) with other processes. Section objects also provide the mechanism by which a process can map a file into its memory address space.
https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/section-objects-and-views

The loop that maps over all sections:
https://github.com/ayyucedemirbas/Windows-Research-Kernel-WRK-/blob/master/WRK-v1.2/base/ntos/mm/creasect.c#L3184-L3411

tests/pefile_test.py
```diff
@@ -658,3 +659,124 @@ def test_checksum(self):
...
+        bytes(dos_header) + b"PE\x00\x00" + file_header + opt_header + section_header
+    )
+    headers = headers.ljust(pointer_to_raw_data, b"\x00")
+
+    # Fill raw section data with a sentinel byte so leakage is obvious
+    section_data = b"\xCC" * size_of_raw_data
+    return bytes(headers) + section_data
+
+
+class Test_memory_mapped_image(unittest.TestCase):
```
j-t-1: `test_virtual_size_less_than_raw_size` and `test_virtual_size_greater_than_raw_size` can be moved to line 345, before `test_relocated_memory_mapped_image`, within class `TestPEFile`.

tests/pefile_test.py
```diff
@@ -658,3 +659,124 @@ def test_checksum(self):
...
+        + struct.pack("<I", size_of_raw_data)  # SizeOfCode
+        + struct.pack("<II", 0, 0)  # SizeOfInitializedData, SizeOfUninitializedData
+        + struct.pack("<I", va)  # AddressOfEntryPoint
+        + struct.pack("<II", va, 0)  # BaseOfCode, BaseOfData
+        + struct.pack("<I", 0x400000)  # ImageBase
+        + struct.pack("<II", section_alignment, 0x200)  # FileAlignment
+        + struct.pack("<HHHHHH", 6, 0, 0, 0, 6, 0)  # OS/Image/Subsystem versions
+        + struct.pack("<I", 0)  # Win32VersionValue
+        + struct.pack("<I", size_of_image)  # SizeOfImage
+        + struct.pack("<I", pointer_to_raw_data)  # SizeOfHeaders
```
j-t-1: SizeOfHeaders: The combined size of an MS-DOS stub, PE header, and section headers rounded up to a multiple of FileAlignment.

hex(64 + 92 + struct.calcsize("8sIIIIIIHHI")) is '0xc4', and `FileAlignment` 0x200.

Therefore, SizeOfHeaders needs changing to the below?
`+ struct.pack("<I", 0x200)  # SizeOfHeaders`

tests/pefile_test.py
binggao1230: Done in 2671cc8 — moved both into `TestPEFile` just before `test_relocated_memory_mapped_image` and dropped the standalone `Test_memory_mapped_image` class.

tests/pefile_test.py
binggao1230: Done in 2671cc8 — set it to `0x200`, the real header size (`0x160`) rounded up to FileAlignment. (The previous `PointerToRawData` value was also a valid SizeOfHeaders, but 0x200 is the minimal/conventional one and clearer here.)

j-t-1: I am glad you said that as confirms my understanding.

Some of the changes to the tests I asked for were ad hoc; in the future, the creation of the PE for the test will be standardized and maybe expanded for code reuse with other tests. Outside the scope of this PR.

@nightlark this PR improves pefile. To give confidence that pefile does memory mapping correctly, I will open an issue about comparing it with other tools. Comparing it to the WRK also is within this comparison.

j-t-1 (approved):

nightlark:
Do we have an issue to remind us to check the behavior of this against how Windows handles the memory mapping? It’s something I want to do before we make a new release.

pefile.py
```diff
@@ -6325,7 +6325,19 @@ def get_memory_mapped_image(self, max_virtual_address=0x10000000, ImageBase=None
...
+            if virtual_size and len(section_data) != virtual_size:
+                # Only Misc_VirtualSize bytes belong in the mapped view.
+                if virtual_size < len(section_data):
+                    # VirtualSize < SizeOfRawData: the extra disk bytes are
+                    # file-alignment padding the OS loader discards.
+                    section_data = section_data[:virtual_size]
+                elif virtual_size > len(section_data):
+                    # VirtualSize > SizeOfRawData: the gap is BSS and must be
+                    # zero-padded in the mapped view.
+                    section_data += b"\x00" * (virtual_size - len(section_data))
```
nightlark, line 6339: This could be really bad for memory usage when pefile is used to analyze malicious files - should test against a file to confirm how bad it is, but if it causes out of memory related crashes or slows things down enough to be unusable we will need to change this to avoid  allocating potential gigabytes of empty bytes before shipping a new release.

nightlark:
I think there are two things from this merged PR that need some additional scrutiny. Depending on what we find it may need to be reverted.

j-t-1:
> Do we have an issue to remind us to check the behavior of this against how Windows handles the memory mapping? It’s something I want to do before we make a new release.

Completely agree about needing to investigate memory mapping.

Here is a first attempt of an issue, and we can put a "help wanted" tag on it.

> The memory mapping needs checking with the results of tools such as x64dbg and YARA-X.
> 
> ```
> # Drivers would also be useful to check
> pe = pefile.PE(name=r"C:\Windows\System32\notepad.exe")
> image = pe.get_memory_mapped_image()
> ```
> 
> If the memory mapping results differ, investigate the reasons why.

How Windows handles memory mapping is the standard so this needs to be incorporated either in this issue or another one.

How about:

I raise the above issue (with any changes you suggest).
You raise an issue about memory mapping matching with Windows and/or how to handle malicious files when memory mapping.

Both tagged with "help wanted".

pefile.py
j-t-1: Given this is one of the usages of pefile, that is valid; I neglected to factor this in.

PR #557, opened 2026-07-13
Fix bare except clauses with specific exception types in pefile.py

koteshyelamati:
Replace bare `except:` clauses with specific exception types in `pefile.py`.

Two bare `except:` clauses handle malformed PE directory structures:
- Access to `last_entry.directory.entries[0].directory.entries` can raise `AttributeError` or `IndexError` → changed to `except (AttributeError, IndexError):`
- Access to `version_entry.data.struct` can raise `AttributeError` or `IndexError` → changed to `except (AttributeError, IndexError):`

These are both "maybe malformed structure, ignore" patterns where the specific exceptions are clear.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -2944,7 +2944,7 @@ class PE:
         fast_load = fast_load if fast_load is not None else globals()["fast_load"]
         try:
             self.__parse__(name, data, fast_load, max_offset)
-        except:
+        except Exception:
             self.close()
             raise
 
@@ -4881,7 +4881,7 @@ class PE:
 
                 try:
                     version_entries = last_entry.directory.entries[0].directory.entries
-                except:
+                except (AttributeError, IndexError):
                     # Maybe a malformed directory structure...?
                     # Let's ignore it
                     pass
--- a/pefile.py
+++ b/pefile.py
@@ -4890,7 +4890,7 @@ class PE:
                         rt_version_struct = None
                         try:
                             rt_version_struct = version_entry.data.struct
-                        except:
+                        except (AttributeError, IndexError):
                             # Maybe a malformed directory structure...?
                             # Let's ignore it
                             pass


j-t-1:
Thanks, clear improvements to the exception handling.

j-t-1:
@nightlark this PR has authored by me, although ideally want to have @koteshyelamati. Would squash and merge have done this? Is there a setting in which the person who did the PR always has their icon and username in the latest commit at https://github.com/erocarrera/pefile?

I will read up on this as my understanding of this process could be better.

nightlark:
Looking at https://github.com/erocarrera/pefile/commit/8996529447e1e5afb4a323bd382bfe5f0ba08f4c, it appears to just be a merge commit (lower right side of it shows 2 parents) with the default text `Merge pull request #NNN from <fork>/<branch>` changed to something else. Merge commits show as authored by whoever clicked the merge button -- but the commits doing the change still show up in git history as being authored by the correct person (https://github.com/erocarrera/pefile/commits/master/?since=2026-07-13&until=2026-07-14).

I think the contributors info showing number of commits by different people doesn't count merge commits.

For squash and merge it shows whoever opened the PR as the author, and any additional authors are only listed by adding `Co-authored-by` lines to end of the commit message for the squashed commit.

koteshyelamati:
Thank you for merging, @j-t-1! And thanks @nightlark for the explanation about merge commit authorship — that's helpful to know.

j-t-1:
Also thanks @nightlark. The git history as being authored by the correct person is most important part, which has happened here.

Squash and merge could be better for us, as having the person who did the commit as the author seems more logical. Also has a cleaner history, with a one-to-one relationship with the PR.

A merge commit would be good for some workflows, like a big change done is sub-components, although pefile as is fairly stable so will occur infrequently. A weakness of merge and commit is for a PR with multiple commits where a later one corrects something in an previous one, the earlier one is still included in the flow of changes.

I slightly prefer squash and merge, but there may be downsides (like you have to edit the message when there are multiple commits and their messages may not stack together well). If you think keep with merge commit, then also fine as keeps consistency.

PR #560, opened 2026-07-15
Fix resource parsing for low-alignment PE images (#465)

### LINKED ISSUE #465: Can't parse resources in a resource-only PE files
Sample file:
[shell32.zip](https://github.com/user-attachments/files/21206173/shell32.zip)

Sample script:
```python
import sys
import pefile

def list_resource_types(dll_path):
    try:
        pe = pefile.PE(dll_path)
        
        if hasattr(pe, 'DIRECTORY_ENTRY_RESOURCE'):
            resource_types = []
            for entry in pe.DIRECTORY_ENTRY_RESOURCE.entries:
                if entry.name is not None:
                    resource_types.append(entry.name)
                else:
                    resource_types.append(f"Type {entry.id}")
            
            print(f"Resource types in {dll_path}:")
            for rt in resource_types:
                print(f"  - {rt}")
        else:
            print(f"No resources found in {dll_path}")
            
    except Exception as e:
        print(f"Error: {e}")

dll_path = sys.argv[1]
list_resource_types(dll_path)

```

Execution:
```
> .\list.py ".\shell32.dll"
No resources found in .\shell32.dll

> .\list.py C:\Windows\System32\shell32.dll
Resource types in C:\Windows\System32\shell32.dll:
  - MUI
  - TYPELIB
  - Type 16
  - Type 24
```

mobinert:
## What

Fixes #465: resources in resource-only PE files (and other low-alignment images) don't parse.

## Why

The sample in the issue is a single-section `.rsrc` DLL with `SectionAlignment == FileAlignment == 0x10`, `PointerToRawData = 0x160`, `VirtualAddress = 0x1a0`.

For low-alignment images (SectionAlignment < the page size) the loader reads a section's data straight from `PointerToRawData` without rounding it down to a 0x200 sector boundary. `SectionStructure.get_PointerToRawData_adj` only accounted for that when `PointerToRawData == VirtualAddress`. Here they differ, so `adjust_PointerToRawData(0x160)` rounded the pointer down to `0`. `get_offset_from_rva(0x1a0)` then resolved to `0` (the MZ header) instead of `0x160`, the resource directory was read from the wrong place, and parsing bailed with:

```
Error parsing the resources directory. The directory contains 65535 entries (>4096)
```

so `DIRECTORY_ENTRY_RESOURCE` was never populated.

## Fix

Use the unrounded `PointerToRawData` for all sections in a low-alignment image, not only when it equals `VirtualAddress` (that case is subsumed). One-line change in `get_PointerToRawData_adj`.

With the fix, the issue's sample resolves `get_offset_from_rva(0x1a0) == 0x160` and its resources parse.

## Testing

- Added `test_low_alignment_section_pointer_to_raw_data`, which builds a minimal low-alignment PE with a resource tree in-memory (no test-corpus file needed). It fails before this change (RVA→offset resolves to 0, no resource types) and passes after.
- Ran the full `erocarrera/pefile-tests` regression suite (748 files) against the change: **no output differences** and no new load failures. Of the 35 low-alignment files in that corpus, none change (their raw pointers are already sector-aligned or equal to `VirtualAddress`), so no `.dmp` fixtures need regenerating.
- `flake8 --select=E9,F63,F7,F82` clean.

Note: I placed the test in `tests/pefile_test.py` alongside the other alignment/`PointerToRawData` tests. Since the tox run currently `--ignore`s that file, happy to port the test to `erocarrera/pefile-tests` instead if you'd prefer it run in CI.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -1139,7 +1139,6 @@ class SectionStructure(Structure):
 
     def get_PointerToRawData_adj(self):
         if self.PointerToRawData_adj is None and self.PointerToRawData is not None:
-            ptrd = self.pe.adjust_PointerToRawData(self.PointerToRawData)
             # When the SectionAlignment is smaller than the native page-size the
             # loader reads a section's data straight from PointerToRawData without
             # rounding it down to a sector boundary (LdrpWx86FormatVirtualImage).
@@ -1149,6 +1148,8 @@ class SectionStructure(Structure):
             # resource directory RVA->offset lookup into the MZ header.
             if self.pe.OPTIONAL_HEADER.SectionAlignment < 0x1000:
                 ptrd = self.PointerToRawData
+            else:
+                ptrd = self.pe.adjust_PointerToRawData(self.PointerToRawData)
             self.PointerToRawData_adj = ptrd
         return self.PointerToRawData_adj
 
--- a/pefile.py
+++ b/pefile.py
@@ -1147,10 +1147,11 @@ class SectionStructure(Structure):
             # 0x160 / VirtualAddress 0x1a0, and rounding that down to 0 sends the
             # resource directory RVA->offset lookup into the MZ header.
             if self.pe.OPTIONAL_HEADER.SectionAlignment < 0x1000:
-                ptrd = self.PointerToRawData
+                self.PointerToRawData_adj = self.PointerToRawData
             else:
-                ptrd = self.pe.adjust_PointerToRawData(self.PointerToRawData)
-            self.PointerToRawData_adj = ptrd
+                self.PointerToRawData_adj = self.pe.adjust_PointerToRawData(
+                    self.PointerToRawData
+                )
         return self.PointerToRawData_adj
 
     def get_VirtualAddress_adj(self):
--- a/pefile.py
+++ b/pefile.py
@@ -1140,13 +1140,15 @@ class SectionStructure(Structure):
     def get_PointerToRawData_adj(self):
         if self.PointerToRawData_adj is None and self.PointerToRawData is not None:
             ptrd = self.pe.adjust_PointerToRawData(self.PointerToRawData)
-            # When the SectionAligment is smaller than the native page-size if the
-            # section’s PointerToRawData and VirtualAddress match, the section's data
-            # will be read at that offset. Implemented in the Window's function:
-            # LdrpWx86FormatVirtualImage.
+            # When the SectionAlignment is smaller than the native page-size the
+            # loader reads a section's data straight from PointerToRawData without
+            # rounding it down to a sector boundary (LdrpWx86FormatVirtualImage).
+            # This isn't limited to sections where PointerToRawData ==
+            # VirtualAddress: resource-only images can have e.g. PointerToRawData
+            # 0x160 / VirtualAddress 0x1a0, and rounding that down to 0 sends the
+            # resource directory RVA->offset lookup into the MZ header.
             if self.pe.OPTIONAL_HEADER.SectionAlignment < 0x1000:
-                if self.PointerToRawData == self.VirtualAddress:
-                    ptrd = self.VirtualAddress
+                ptrd = self.PointerToRawData
             self.PointerToRawData_adj = ptrd
         return self.PointerToRawData_adj
 
--- a/tests/pefile_test.py
+++ b/tests/pefile_test.py
@@ -10,44 +10,6 @@ import pefile
 REGRESSION_TESTS_DIR = "tests/test_files"
 
 
-def _low_alignment_resource_pe():
-    """Minimal PE32 with SectionAlignment == FileAlignment == 0x10 and a single
-    .rsrc section whose PointerToRawData (0x160) differs from its VirtualAddress
-    (0x1a0), mirroring the resource-only DLL reported in issue #465."""
-    va, ptr_raw = 0x1A0, 0x160
-
-    # resource tree, offsets relative to the start of .rsrc (RVA 0x1a0)
-    res = b""
-    res += struct.pack("<IIHHHH", 0, 0, 0, 0, 0, 1)  # type dir, one id entry
-    res += struct.pack("<II", 3, 0x80000000 | 0x18)  # type 3 -> name dir
-    res += struct.pack("<IIHHHH", 0, 0, 0, 0, 0, 1)  # name dir
-    res += struct.pack("<II", 1, 0x80000000 | 0x30)  # name 1 -> lang dir
-    res += struct.pack("<IIHHHH", 0, 0, 0, 0, 0, 1)  # lang dir
-    res += struct.pack("<II", 0x409, 0x48)           # lang -> data entry
-    res += struct.pack("<IIII", va + 0x58, 4, 0, 0)  # data entry
-    res += b"TEST"
-
-    dos = b"MZ" + b"\x00" * 0x3A + struct.pack("<I", 0x40)
-    coff = struct.pack("<HHIIIHH", 0x14C, 1, 0, 0, 0, 0xE0, 0x2102)
-    size_of_image = (va + len(res) + 0xF) & ~0xF
-    opt = struct.pack("<HBBIIIIII", 0x10B, 0, 0, 0, 0, 0, 0, 0, 0)
-    opt += struct.pack("<III", 0x400000, 0x10, 0x10)
-    opt += struct.pack("<HHHHHH", 4, 0, 0, 0, 4, 0)
-    opt += struct.pack("<I", 0)
-    opt += struct.pack("<II", size_of_image, ptr_raw)
-    opt += struct.pack("<IHH", 0, 2, 0)
-    opt += struct.pack("<IIII", 0, 0, 0, 0)
-    opt += struct.pack("<II", 0, 16)
-    dirs = [(0, 0)] * 16
-    dirs[2] = (va, len(res))  # resource directory
-    for rva, sz in dirs:
-        opt += struct.pack("<II", rva, sz)
-    section = struct.pack(
-        "<8sIIIIIIHHI", b".rsrc\x00\x00\x00", len(res), va, len(res), ptr_raw, 0, 0, 0, 0, 0x40000040
-    )
-    return dos + b"PE\x00\x00" + coff + opt + section + res
-
-
 class TestPEFile(unittest.TestCase):
     maxDiff = None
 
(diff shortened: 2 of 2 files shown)


j-t-1:
Thanks, this PR fixes the issue. 

As well as resource-only PEs, `self.pe.OPTIONAL_HEADER.SectionAlignment < 0x1000` is going to be met when they are system files and drivers? I do not know if drivers are affected, and which system files would be. Those files may still need the `self.PointerToRawData == self.VirtualAddress`? If necessary, I can ask this as an issue.

I was thinking adding a `PE.is_dll` check to this PR but that increases coupling too much. At the Section level you cannot tell if it is from a resource-only PE file?

Does moving `_low_alignment_resource_pe` below `TestPEFile` work (so when reading the tests you go straight to `TestPEFile`)?

For the note, the changes you made here are not being actually run by the tests above as they use `erocarrera/pefile-tests`? Until you said this, I thought they were both run! Will have to look at this (in general) and clarify how we are using the tests here and in the other repo.

mobinert:
Thanks for the careful review.

On the drivers / system files question: the rounding I removed was only ever correct for normal-alignment images. When `SectionAlignment < 0x1000` the loader requires `FileAlignment == SectionAlignment` (so it's also below 0x200), and it maps the image flat, reading each section straight from `PointerToRawData`. There's no sector rounding in that path, so the unrounded `PointerToRawData` is what the loader uses for every section, not only the ones where `PointerToRawData == VirtualAddress`.

That's why I don't think drivers or system files are at risk here:

- Anything with `SectionAlignment >= 0x1000` never enters this branch, so ordinary drivers and system DLLs are untouched.
- For a genuinely low-alignment file of any kind, flat mapping means the unrounded pointer is still the correct offset.
- The old `PointerToRawData == VirtualAddress` case is a strict subset of the new behaviour: when they're equal, the unrounded `PointerToRawData` already equals `VirtualAddress`, so those files resolve identically.

So the only files whose offsets change are low-alignment sections where `PointerToRawData != VirtualAddress` and the pointer wasn't already sector-aligned, and those were being read from the wrong place before (rounded down, in the #465 sample all the way to 0, i.e. the MZ header). Nothing that parsed correctly under the old rounding regresses. That lines up with the corpus run too: no output diffs across the 748 files, and the 35 low-alignment ones don't change. I don't think a separate issue is needed, but I'm happy to write one up for the record if you'd like it tracked.

On telling at the Section level whether it came from a resource-only PE: you can't, and I'd argue it shouldn't matter. The flat-mapping rule is a property of the image's alignment, not of it being a DLL or resource-only. Resource-only DLLs are just where it shows up most often because they tend to use tiny alignments. `SectionAlignment < 0x1000` is exactly the right signal to key off. An `is_dll` / resource-only check would add coupling and be slightly wrong, since it would miss non-DLL low-alignment images while not changing the resource-only result anyway. So I'd keep it as is.

I moved `_low_alignment_resource_pe` below `TestPEFile` as you suggested, so the class is what you hit first reading top-down. It still resolves fine since the method only calls it at runtime.

On the tests: sorry for the surprise there. The `tox` run `--ignore`s `tests/pefile_test.py`, so what actually executes against a checkout is the `erocarrera/pefile-tests` corpus, and this new test lives in `pefile_test.py` so it validates locally but won't run in that CI path. If you'd rather have it gated in CI I can port it over to `pefile-tests`; otherwise I'm happy to leave the wider clarification to you.

pefile.py
```diff
@@ -1140,13 +1140,15 @@ def __init__(self, *args, **kwargs):
     def get_PointerToRawData_adj(self):
         if self.PointerToRawData_adj is None and self.PointerToRawData is not None:
             ptrd = self.pe.adjust_PointerToRawData(self.PointerToRawData)
```
j-t-1: Can this be under an else clause, below the last line you modify? Do you think this makes the function more readable? The two branches could then assign to `self.PointerToRawData_adj`, avoiding the intermediate `ptrd`.

j-t-1:
Thanks for this. I think you could be correct.

Is the reason that VirtualAddress (RVA) is sometimes the same as the PointerToRawData is it simplifies loading somehow? So it is common for this case, hence why the previous check?

Either inside or outside of this PR, a link to relevant information on LdrpWx86FormatVirtualImage could be useful.

mobinert:
Moved the `adjust_PointerToRawData` call into an `else` so it's only computed on the normal-alignment path (c01f9dd).

On why `PointerToRawData` so often equals `VirtualAddress` here — it's actually required by the spec for these files, which is what the old check was leaning on. From the PE format docs, "Section Data":

> There are additional restrictions on image files if the SectionAlignment value in the optional header is less than the page size of the architecture. For such files, the location of section data in the file must match its location in memory when the image is loaded, so that the physical offset for section data is the same as the RVA.

So a well-formed low-alignment image is mapped flat (file offset == RVA), and `PointerToRawData == VirtualAddress` just falls out of that. That's the common case the old `== VirtualAddress` branch covered.

The #465 sample doesn't hold to that constraint: `PointerToRawData` is `0x160` while `VirtualAddress` is `0x1a0`. Windows still loads it because the section header's `PointerToRawData` is what says where the bytes actually are on disk, and for low-alignment images it reads them from there without the `0x200` rounding. Reading the resource directory from `0x160` gives a clean icon resource tree; the rounded-to-0 path was landing in the MZ header. So the change is really just generalizing the existing special case from "equal" to "any low-alignment section".

Reference: [PE Format — Section Data](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-data) (the `SectionAlignment` < page size note). The pe-bear issue already referenced in the tests (hasherezade/pe-bear#11) is the same corner from that side.

mobinert:
Also dropped the intermediate `ptrd` — both branches now assign to `self.PointerToRawData_adj` directly (55c08a1). Agreed it reads better that way.

j-t-1 (approved):

j-t-1:
> Note: I placed the test in tests/pefile_test.py alongside the other alignment/PointerToRawData tests. Since the tox run currently --ignores that file, happy to port the test to erocarrera/pefile-tests instead if you'd prefer it run in CI.

Options:

1. Move all tests to erocarrera/pefile-tests.
2. Move test code from erocarrera/pefile-tests to here, so that repo is just for PE file storage.
3. Keep any tests not relying on actual PE files in this repo.

Option 3. seems a good compromise, which is what this PR does. For tests/pefile_test.py not running in CI, we could have an action or modify pyproject.toml.

@nightlark I think you understand this better, what would you advise?

mobinert:
Thanks for approving!

Between the two I'd lean towards option 2 — keep the test code here in `pefile` and let `pefile-tests` be just the PE corpus. The main reason is exactly what we ran into on this PR: when the code and its test live in the same repo, a change to `pefile.py` and its test go in one PR and get gated together. As it stands a fix here can't bring its own test into CI without a second PR to the other repo, which is easy to forget. The heavy binary corpus staying in a separate data repo (pulled in CI the way it is now) still keeps this repo light.

Option 1 works too, but then every code change that needs a test becomes a two-repo dance and the test can drift from the code it covers.

Either way I'm happy to help with the move. For this PR specifically the new test is self-contained (it builds the PE in memory, no corpus file), so under option 2 it just runs here as-is; under option 1 I can port it over to `pefile-tests`.

j-t-1:
Thinking some more, because the test repo is a fork of a fork (although now archived), we should probably not change this without getting the upstream owners feedback. If changing things meant they would be less likely to contribute to the tests repo corpus that would be a negative. Moving all the test code here would mean that the test repo would be more general (not tied to pefile as much) and may encourage people to use it as a more general store of interesting PE files (good for us as well), but moving the test code does not seem ours to move without their feedback. We could more easily move the pefile test code to the test repo, simplifying having all test stuff in one location but has the negatives you mention.

I think this discussion would be in a new issue, and link in the owners of the two upstream forks and @erocarrera for their opinions. The new release will have a lot of changes already so could wait until the new release and then raise an issue.

nightlark:
I'd lean towards 2 or 3, with pefile-tests being focused on a corpus of PE test files rather than containing the test scripts run as part of pefile CI jobs.

https://github.com/pombredanne/pefile-tests/pull/1 has been open for over 3 years now and not merged upstream; I think at this point erocarrera's fork of the upstream is effectively the main repo by virtue of the other two not having any updates in 4 and 9 years. Since the primary purpose is as a corpus of files for testing pefile, I think it makes sense to treat it as independent of whatever the upstream repositories shown by GitHub are doing (+ we don't want pefile CI tests to break due to a change in a repo that we have no control over).

I agree with waiting until we make a new release. Currently busy getting ready for presenting at conferences, but I'll have more time around August to dig more into the open PRs/issues that require a deeper dive to understand. Maybe aiming for an end of August/early September for a new release?

j-t-1:
Also prefer option 2 or 3, with 2 slightly better. Alongside the reasons given by @mobinert, having the test scripts in one place would make it easier to get an overview of what we are and are not yet testing.

Good points about the upstream repo. I thought there was more code in pefile-tests, but is mainly
https://github.com/erocarrera/pefile-tests/blob/master/tests/pefile_test.py. Overall, moving it here is easier than my previous comment suggested. If we go this route, opening an issue and linking the upstream owner would still be a good to do.

Late August or early September would be really good.

j-t-1:
> Between the two I'd lean towards option 2 — keep the test code here in `pefile` and let `pefile-tests` be just the PE corpus. The main reason is exactly what we ran into on this PR: when the code and its test live in the same repo, a change to `pefile.py` and its test go in one PR and get gated together. As it stands a fix here can't bring its own test into CI without a second PR to the other repo, which is easy to forget. The heavy binary corpus staying in a separate data repo (pulled in CI the way it is now) still keeps this repo light.

For these reasons, I agree and think we do this. @nightlark I know you are busy short term, but okay with you?

> Either way I'm happy to help with the move. For this PR specifically the new test is self-contained (it builds the PE in memory, no corpus file), so under option 2 it just runs here as-is; under option 1 I can port it over to `pefile-tests`.

@mobinert are you still able to do this? Although I said this could be after the new release, there could be advantages to getting the tests in order (including having them all running) prior to the new release, as there could be a lot of new issues afterwards. Let's see how things progress, and regardless the changes will be integrated at latest shortly after the new release, which is soon.

@pombredanne gave a really positive response in #584 and may be able to help if we run into issues.

mobinert:
Yes, I'm still able to help with this. I'll review the current state in #584 and follow up there with a focused plan before making any changes.

PR #583, opened 2026-07-30
Consolidate to using SHA-256 for sample files

j-t-1:
This was done using VirusTotal.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -994,8 +994,8 @@ class Structure:
             data = data[: self.__format_length__]
 
         # Some malware have incorrect header lengths.
-        # Fail gracefully if this occurs
-        # Buggy malware: a29b0118af8b7408444df81701ad5a7f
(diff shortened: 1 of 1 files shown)


j-t-1:
Two sample files are not available on VirusTotal:

https://github.com/erocarrera/pefile/blob/ac56742648ada10fa2c0d99433a322d471ddb4c2/pefile.py#L3216-L3219

https://github.com/erocarrera/pefile/blob/ac56742648ada10fa2c0d99433a322d471ddb4c2/pefile.py#L3317-L3326

Both were in the initial commit of pefile.

@nightlark, suggest removal of lines 3217 and 3319.

nightlark:
Maybe @erocarrera has the files with those hashes in the encrypted set of test files?

Another idea is we could create some synthetic test files for these code paths. Or maybe find other files on VT that are similar.

PR #595, opened 2026-08-10
Refactor set_format

j-t-1:
Because it is data, not formatting information, `__unpacked_data_elms__ ` is not a good fit for set_format. It is also redundant, and can be created from other return values of the function.

### DIFF
--- a/pefile.py
+++ b/pefile.py
@@ -29,7 +29,7 @@ import string
 import struct
 import time
 import uuid
-from collections import Counter, defaultdict
+from collections import Counter, namedtuple, defaultdict
 from functools import lru_cache, wraps
 from hashlib import md5, sha1, sha256, sha512
 
@@ -919,12 +919,9 @@ def set_format(format):
 
     __format_length__ = struct.calcsize(__format_str__)
 
-    return (
-        __format_str__,
-        __format_length__,
-        __field_offsets__,
-        __keys__,
-    )
+    Format = namedtuple('Format', '__format_str__, __format_length__, __field_offsets__, __keys__')
+
+    return Format(__format_str__, __format_length__, __field_offsets__, __keys__)
 
 
 class Structure:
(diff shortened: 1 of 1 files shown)


j-t-1:
`__unpacked_data_elms__` is removed from `set_format` for these reasons:

- It is data, not formatting information.
- It is an extra parameter, that confusingly is not used (it will be replaced when the actual data is obtained elsewhere).
- `set_format` creates a list of `None` objects, and we know the length of this from other returned values.
- Because it is cached it adds memory use.

j-t-1:
@nightlark I will squash and merge this if the change is okay with you and passes code review.

nightlark:
The caching could potentially use less memory than with this change.

The main thing that comes to mind is that I am curious how much of a breaking change this might be for other software that uses pefile, will look around some.

j-t-1:
We could do a deprecate with replacement?

In this code the Structure is created and only used for the `sizeof`:

https://github.com/erocarrera/pefile/blob/65b8f0cd307d55611553abffdcb935ce1ce2d114/pefile.py#L4005-L4010

Which is just returning an output of `set_format`:

https://github.com/erocarrera/pefile/blob/65b8f0cd307d55611553abffdcb935ce1ce2d114/pefile.py#L988-L990

I think it could be better calling `set_format` directly here. If it was called `get_format` it would a better name for the API. (`get_format` would be the replacement in the deprecation of `set_format`.) A negative is the `sizeof` name is self-documenting, whereas selecting an index from a tuple is not.

j-t-1:
Further change.

**Return a named tuple from these functions**
The elements can then be obtained by name and without having to create a`Structure` or `StructureWithBitfields`.

@Auscitte this PR copies the output of the first four items of `set_bitfields_format` as that is a more logical order. Would you be able to help and provide feedback on the changes in this PR?

Issue #371, opened 2023-03-29
Feature request: function for checking of validity of PE files


TheMythologist:
Hi, I'm looking through the documentation and I don't think there is currently a way to check if a file / bytes is a valid PE file, other than via the following `try...except` excerpt:
```python
import pefile

try:
    pefile.PE(filename)
    print("Valid PE file")
except pefile.PEFormatError:
    print("Invalid PE file")
```

Looking through the functions available, I see the function `is_valid` in `peutils.py`, but it currently does nothing (simply a `pass` statement).

Would it be alright for me to create a pull request, targeting the updating of this function? This will help to implement the functionality of checking if a file is valid PE file, similar to the `zipfile` standard library's function `is_zipfile`.

erocarrera:
Hi, if you have a use for such functionality, feel free to send a PR. 
I would avoid the "is valid PE" naming. I feel like saying that a PE is valid implies that it is a correct, well-formed, functioning PE, which is a hard thing to say for sure. I would limit it to a `is_pe_file()` method that attempts a light-weight parse (with `fast_load=True`). That would differentiate between a parseable PE and non-PE data.

j-t-1:
The YARA logic is a slightly better fit, still not that strict but probably okay for an initial working function for pefile.

```
def is_pe_file(pe):
    """The paper "Lost in the Loader: The Many Faces of the Windows PE File
    Format" identifices three broad operations performed by software on the PE
    format: structrual checks, compliance checks, and memory mapping.

    "As a rule of thumb, structural checks verify if a file is well formed, while
    compliance checks test whether the extracted information is valid and can be
    used to perform the task of the software.  However, the boundary between
    structural and compliance checks is not always clear because of the
    ambiguities in the specifications ofthe PE Format that fails to declare the
    structure of a PE executablein a formal way. Despite this difficulty, we
    believe the distinction among the two types of checks can ease the
    discussions of the various aspects of our work."    

    This function uses the YARA checks, for rudimentary structural and compliance checks.
    https://github.com/eurecom-s3/loaders-models/blob/d29d62bff9bff7f53f086f661b00168ffc082227/avs/yara/yara.lmod

    Note: The Windows loader does more complex checks and includes all three
    types of checks, and whether a PE file gets loaded varies between different
    versions.
    """
    # IMAGE_DOS_SIGNATURE is checked during initial parsing, redundantly included for completeness
    from pefile import IMAGE_DOS_SIGNATURE
[...]

PR #604, opened 2026-09-08
Add basic structural and compliance check

j-t-1:
Checks IMAGE_DOS_SIGNATURE, IMAGE_NT_SIGNATURE and whether the section headers specify raw data that is within the size of the file.

It is a subset of the few constraints YARA does when parsing PE files, copied from
https://github.com/eurecom-s3/loaders-models/blob/d29d62bff9bff7f53f086f661b00168ffc082227/avs/yara/yara.lmod


j-t-1:
Closes #371.

PR #612, opened 2026-09-12
Change first element retrieval

j-t-1:
Fetch the first element instead of creating an unnecessary new list. This was identified and fixed using Ruff.
RUF015: unnecessary-iterable-allocation-for-first-element.


j-t-1:
> This rule's fix is marked as unsafe, as migrating from (e.g.) list(...)[0] to next(iter(...)) can change the behavior of your program in two ways:
> 
> 1. First, all above-mentioned constructs will eagerly evaluate the entire collection, while next(iter(...)) will only evaluate the first element. As such, any side effects that occur during iteration will be delayed.
> 2. Second, accessing members of a collection via square bracket notation [0] or the pop() function will raise IndexError if the collection is empty, while next(iter(...)) will raise StopIteration.

https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/

I do not think these apply to the changes.

nightlark:
I agree with the 1st of those two things not applying in this case, and probably the 2nd one - but mostly because if StopIteration being raised were a problem, then accessing array index zero would’ve also been a problem. Would need to take a look at where the FileInfo.Var entries are being created to make sure there’s never a way for them to have an empty entry.

